PR #1, opened 2020-12-01
Initial work

evansd:
A first working version of the `opensafely` cli tool.


README.md
```diff
@@ -0,0 +1,56 @@
+# OpenSAFELY Command Line Interface
+
+This tool is part of the [OpenSAFELY](https://www.opensafely.org)
+project.
+
+To use it you will need:
+ * Python version 3.7 or above
```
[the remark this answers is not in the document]
evansd: We had to remove 3.8-specific syntax from the original cohortextractor because people didn't have that installed locally and we were trying to minimise friction. I guess eventually we'll be shipping our own binary and this will be irrelevant, but right now I want to make sure that this works in whatever environment people have currently been using.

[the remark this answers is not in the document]
ghickman: 👌 

Not relevant to this PR I think but probably best to write that down somewhere?

ghickman (approved):

bloodearnest (approved):
LGTM

Intrigued by pulling job-runner via release rather than git hash, but I supose it'll be pypi package once v2 is landed.

evansd:
Yeah, pulling direct from github is just a stopgap until the v2 branch gets merged.

It _is_ via a hash though, not a release (github.com/opensafely/job-runner/archive/**86d2127**.tar.gz). It's just using a tarball rather than a git clone.

README.md
```diff
@@ -0,0 +1,29 @@
...
+To minimise the possibility for installation issues this package vendors
+all its dependencies under the [opensafely._vendor](./opensafely/_vendor)
+namespace using the [vendoring](https://pypi.org/project/vendoring/) tool.
+
+Note that while the `opensafely-jobrunner` package technically depends
+on `requests` this is only used by the `jobrunner.sync` module which is
+never invoked by this interface (as smoketests will confirm).
+
+To update the vendored dependencies first update the versions in
+[vendor.txt](./vendor.txt).
```
sebbacon, line 44: What is the story for deciding what to vendor? If a new dependency is added to jobrunner which needs vendoring, I guess we're relying on the smoke tests to tell us that it's missing? If so, worth saying so here? Perhaps also saying "run tests locally aftering running vendoring tool"?

README.md
```diff
@@ -1,6 +1,33 @@
...
+ * Python version 3.7 or above
+ * git
+ * Docker
+
+Once these are installed you can install the tool itself with:
+```
+pip install opensafely
+```
+
+At present there is one command, which can be used to run actions in an
```
sebbacon, line 16: I think this is begging to become stale! The documentation *is* the argparse help.
 Better to say "run `opensafely --help` to see permitted arguments"?


https://github.com/DataDog/mkdocs-click

sebbacon (approved):
Looks good.

One suggestion: consider using `click` rather than `argparse`?

Reasons:
1. I can imagine us going doing the subcommand route, which is much easier in argparse (but YAGNI)
2. Ansi colours and other terminal-friendly features built-in (but YAGNI)
3. Can [integrate into our docs automatically](https://github.com/DataDog/mkdocs-click)

README.md
evansd: We're now vendoring all the dependencies so I think this question goes away.

README.md
evansd: :+1: quite right

evansd:
@ghickman @bloodearnest @sebbacon Thanks for the reviews. Can I trouble you for a second round?

I realised that we needed to support the `update_codelists` command from `cohortextractor` which I've copied over. This meant that we needed `requests` so in one sense it makes our vendoring story simpler and we're not just vendoring our entire dependency tree. However it turns out `requests` does various bits of import chicanery which make vendoring it a bit more tricky. Fortunately the issues all show up immediately at import time so the tests which builds, installs and runs the package catches them for us.

README.md
```diff
@@ -0,0 +1,75 @@
...
+all its dependencies under the [opensafely._vendor](./opensafely/_vendor)
+namespace using the [vendoring](https://pypi.org/project/vendoring/) tool.
+
+This brings its own complexities (particularly around the `requests`
+package) but they are at least complexities which show up at development
+time rather than on some researcher's mysteriously broken Python
+installation.
+
+The tool makes the process relatively painless. There are a few things
+workaround we needed to apply which are all configured in
```
ghickman, line 48: I'm struggling to parse the second sentence here!

ghickman (changes requested):
Looking good!  Just that one comment to fix.

Something to consider for later: pushing the package contents into a src directory can make testing installation of a package easier.

ghickman (approved):
👌

PR #4, opened 2020-12-04
fix: Actually maybe publish this time

evansd:
It turns out clicking Update Selection is not sufficient to change which
repos have access to an org secret; you also need to click a separate
Save button.


inglesp (approved):
vg

PR #7, opened 2020-12-07
Move `update_codelists` to `codelists update`

evansd:
This is in preparation for adding more codelist related commands.

Technically this is a major version bump but as no one is officially
using this tool yet we can stick with 1.x for now.

Also script up the dependency updating process and run it.


opensafely/_vendor/certifi/cacert.pem
```diff
@@ -4604,3 +4282,44 @@ AZKXRRJ+oPM+rRk6ct30UJMDEr5E0k9BpIycnR+j9sKS50gU/k6bpZFXrsY3crsC
 MGclCrEMXu6pY5Jv5ZAL/mYiykf9ijH3g/56vxC+GCsej/YpHpRZ744hN8tRmKVu
 Sw==
 -----END CERTIFICATE-----
+
```
CarolineMorton, line 4285: Should this certificate be visible or can it be stashed within the repo settings as TOKEN. I've not done this with github actions but have with Travis and it seemed ok?

CarolineMorton (approved):
Looks fine - not sure if we want the `.pem` published?

opensafely/_vendor/certifi/cacert.pem
evansd: Yes, we do need this. There are public root certificates which the `requests` library uses for verifying HTTPS connections. As we're vendoring in all our dependencies these get pulled in as well.

Always good to query anything that looks like a key though!

CarolineMorton: cool thanks for the explanation

PR #17, opened 2020-12-18
Add `codelists check` command

### LINKED ISSUE #12: Add `codelists check` command
This will replace what we currently do in tests which is to update all codelists and then ask git if anything has changed.

Following the issue we had on Friday where a small change in OpenCodelist's output caused all tests to fail, I propose a different approach to checking codelists which doesn't rely on an external service.

The idea would be that when we run `codelists update` to download the files specified in `codelists.txt` we'd keep a record of exactly what we downloaded in a manifest file called (something like) `codelists.json`. This would record the identifier used in `codelists.txt`, the URL we downloaded from, the download date, and the hash of the file we downloaded.

The `check` command could then ensure that these codelists hadn't been modified, or any extra ones added, without needing to talk to OpenCodelists. The extra information (dates, URLs) in the download manifest might potentially be useful for debugging and audit purposes.

This, as @inglesp pointed out, would roughly mirror the `requirements.in` -> `requirements.txt` relationship in pip-tools.

evansd:
This adds an `opensafely codelists check` command which verifies that the CSVs in your `codelists` folder match what's specified in `codelists.txt`. It does this by recording what files have been downloaded (and their hashes) in a `codelists.json` file.

There's some temporary code which spots that it's being run in a Github Action and falls back to using the old verification method (which checks against OpenCodelists) so that existing repos don't fail immediately under the new method.

Closes #12

### DIFF
--- a/opensafely/codelists.py
+++ b/opensafely/codelists.py
@@ -75,6 +75,7 @@ def update(codelists_dir=None):
         new_files.add(codelist.filename)
         key = str(codelist.filename.relative_to(codelists_dir))
         manifest["files"][key] = {
+            "id": codelist.id,
             "url": codelist.url,
             "downloaded_at": f"{datetime.datetime.utcnow()}Z",
             "sha": hash_bytes(response.content),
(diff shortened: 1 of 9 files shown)


opensafely/codelists.py
```diff
@@ -30,27 +54,162 @@ def main():
...
-    for file in old_files - new_files:
-        print(f"Deleting {file.name}")
-        file.unlink()
+
+def preserve_download_dates(manifest, old_manifest_file):
+    """
+    If file contents are unchanged then we copy the original download date from
+    the existing manifest. This makes the update process idempotent and
+    prevents unnecessary diff noise.
+    """
```
inglesp, line 188: Very nice.

opensafely/codelists.py
```diff
@@ -30,27 +54,162 @@ def main():
...
+            f"'{CODELISTS_DIR}' folder."
+        )
+    for filename, details in manifest["files"].items():
+        csv_file = codelists_dir / filename
+        sha = hash_bytes(csv_file.read_bytes())
+        if sha != details["sha"]:
+            exit_with_prompt(
+                f"A CSV file seems to have been modified since it was downloaded:\n"
+                f"  {CODELISTS_DIR}/{filename}\n"
+            )
```
inglesp, line 133: In each of these error cases, can you give more details (eg which files have been deleted?) and provide exact instructions for the command to run?

inglesp (approved):
:+1:

One suggested change.

tests/test_codelists.py
```diff
@@ -34,3 +38,44 @@ def test_codelists_update(tmp_path, requests_mock):
...
+    assert manifest["files"].keys() == {
+        "project123-codelist456.csv",
+        "project123-codelist098.csv",
+    }
+
+
+@pytest.fixture
+def codelists_path(tmp_path):
+    shutil.copytree(
+        Path(__file__).parent / "fixtures/codelists", tmp_path / "codelists"
```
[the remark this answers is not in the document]
evansd: Good point. As it happens, I think Windows does the right thing with forward slashes but agree it would be better not to rely on this. This has also prompted me to check my belief that the tests were run on Windows in CI which isn't actually true, so I should fix that as well.

opensafely/codelists.py
evansd: The command to run is the same in every case (`opensafely codelists update`) so I made the `exit_with_prompt` function to add this to the bottom of every error message.

As to giving details of the specific files involved, I just got slightly lazy as it's a two-way set diff, two times and ultimately the corrective action is the same anyway. But you're right, it would be helpful to see what exactly has been changed.

PR #24, opened 2021-01-18
fix: Support user-namespaced codelists

inglesp:
Codelists owned by users are at codelists.opensafely.org/codelist/user/[username]/[codelist]/[version].


evansd (approved):
Looks good, thanks!

PR #31, opened 2021-02-19
Add a 'pull' subcommand, which updates docker images.

bloodearnest:
By default it will only update images that are already present locally,
but you can force all images with --force. You can also pull a specific
image.

### DIFF
--- a/opensafely/pull.py
+++ b/opensafely/pull.py
@@ -1,0 +1,57 @@
+import subprocess
+import sys
+
+
+DESCRIPTION = (
+    "Command for updating the docker images used to run OpenSAFELY studies locally"
+)
+REGISTRY = "ghcr.io/opensafely-core"
+IMAGES = ["r", "python", "jupyter", "stata-mp"]
+
+
+def add_arguments(parser):
+    choices = ["all"] + IMAGES
+    parser.add_argument(
+        "image",
+        nargs="?",
+        choices=choices,
+        help="OpenSAFELY docker image to update (default: all)",
+        default="all",
+    )
+    parser.add_argument(
+        "--force",
+        action="store_true",
+        help="Update docker images even if not present locally",
+    )
+
+
+def main(image, force):
+    if image == "all":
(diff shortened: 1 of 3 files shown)


bloodearnest:
Fixes #31

ghickman (approved):

opensafely/pull.py
```diff
@@ -0,0 +1,73 @@
+import subprocess
+import sys
+
+
+DESCRIPTION = (
+    "Command for updating the docker images used to run OpenSAFELY studies locally"
+)
+REGISTRY = "ghcr.io/opensafely-core"
+IMAGES = ["r", "python", "jupyter", "stata-mp"]
```
evansd, line 9: Should we be getting these out of jobrunner instead? It would be very easy to add a new image there and forget to support it here.

opensafely/pull.py
```diff
@@ -0,0 +1,73 @@
...
+        sys.exit(exc.stderr)
+
+
+def get_local_images():
+    ps = subprocess.run(
+        ["docker", "image", "ls", "--format='{{.Repository}}'"],
+        check=True,
+        capture_output=True,
+    )
+    return set(ps.stdout.split("\n"))
```
evansd, line 64: Might be better off using `splitlines()` here for cross-platform purposes. If we end up with `\r` in the strings then the `tag in local_images` check will give false negatives.

evansd (commented):
This is so great, thanks for doing it. Two small comments that might be worth thinking about.

opensafely/pull.py
bloodearnest: Oh, good idea.

opensafely/pull.py
bloodearnest: aha, yes, will do

opensafely/pull.py
bloodearnest: Hoo boo - I forgot about cohortextractor!

Very good call :D

PR #32, opened 2021-02-19
fix: fix text encoding bug in new pull command

bloodearnest:


ghickman (approved):

opensafely/pull.py
```diff
@@ -57,11 +57,12 @@ def main(image, force):
 
 def get_local_images():
     ps = subprocess.run(
-        ["docker", "image", "ls", "--format='{{.Repository}}'"],
+        ["docker", "image", "ls", "--format={{.Repository}}"],
         check=True,
         capture_output=True,
     )
-    return set(ps.stdout.split("\n"))
+    lines = [line for line in ps.stdout.decode('utf8').split("\n") if line.strip()]
```
evansd, line 64: Ah, I see this invalidates my comment in the previous PR. Worth noting you can use `text=True` as an arg to `subprocess.run` to get string rather than bytes output.

PR #38, opened 2021-02-26
Fix `upgrade` command on Windows

### LINKED ISSUE #36: `upgrade` command reports error on Windows (but seems to work anyway)
Running `opensafely upgrade` on Windows reports an error:
```console
$ opensafely --version
Warning: there is a newer version of opensafely available - please run 'opensafely upgrade' to update to 1.6.2

opensafely v1.6.0

$ opensafely upgrade
Collecting opensafely==1.6.2
  Using cached opensafely-1.6.2-py3-none-any.whl (699 kB)
Installing collected packages: opensafely
  Attempting uninstall: opensafely
    Found existing installation: opensafely 1.6.0
    Uninstalling opensafely-1.6.0:
      Successfully uninstalled opensafely-1.6.0
ERROR: Could not install packages due to an OSError: [WinError 5] Access is denied: 'C:\\Users\\dave\\AppData\\Local\\Temp\\pip-uninstall-qxxt6o_c\\opensafely.exe'
Consider using the `--user` option or check the permissions.

Command '['pip', 'install', '--upgrade', 'opensafely==1.6.2']' returned non-zero exit status 1.

$ opensafely --version
opensafely v1.6.2
```

Presumably this is because Windows won't let you write to the running executable. However the update still works anyway, because the entrypoint executable is just a thin wrapper to run Python in the correct venv and pointing to the correct module, and so it doesn't vary at all between versions.

So in one sense this doesn't really matter much. But the error message will no doubt cause confusion. And I suppose we can't guarantee that there won't be cases in future where updating the entrypoint _is_ important.

I think our options are (in decreasing order of laziness):

 1. Do nothing.
 
 2. Print a notice (either before or after pip runs) saying that the error can be ignored.
 
 3. Filter the output from `pip` and remove this specific warning.
 
 4. Try to run `pip` in a process that runs after the parent has executed. I don't know enough about the Windows process model to know how do-able this is. But perhaps if we invoke the subprocess like [this](https://stackoverflow.com/a/12854376) and then exit immediately, it will still work?

evansd:
Also fixes a couple of edge case bugs and smoketests the upgrade command.

Closes #36

### DIFF
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -52,7 +52,7 @@ def main():
     # following recent documentation but has an old opensafely installed,
     # there's some hint as to why their invocation is failing before being told
     # by argparse.
-    if sys.argv[1] != "upgrade":
+    if len(sys.argv) == 1 or sys.argv[1] != "upgrade":
         upgrade.check_version()
 
     args = parser.parse_args()
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -59,7 +59,7 @@ def main():
     kwargs = vars(args)
     function = kwargs.pop("function")
     success = function(**kwargs)
-    sys.exit(0 if success else 1)
+    sys.exit(0 if success is not False else 1)
 
 
 if __name__ == "__main__":
--- a/opensafely/upgrade.py
+++ b/opensafely/upgrade.py
@@ -79,6 +79,8 @@ def get_latest_version(force=False):
 
 
 def comparable(version_string):
+    if version_string == "not-from-a-package":
+        return (0,)
     try:
         return tuple(int(s) for s in version_string.split("."))
     except Exception:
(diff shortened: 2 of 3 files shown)


evansd:
:disappointed: This worked so well on my local Windows VM...

opensafely/__init__.py
```diff
@@ -52,14 +52,14 @@ def show_help(**kwargs):
     # following recent documentation but has an old opensafely installed,
     # there's some hint as to why their invocation is failing before being told
     # by argparse.
-    if sys.argv[1] != "upgrade":
+    if len(sys.argv) < 2 or sys.argv[1] != "upgrade":
```
bloodearnest, line 55: Minor nitpick - I think `if len(sys.argv) == 1` is easier to understand, perhaps? it can never be len 0, right?

bloodearnest: Oh nevermind, I get it

bloodearnest (approved):

opensafely/__init__.py
evansd: Yep, fair point. I've amended.

PR #49, opened 2021-06-01
feat: add --preseed option to pull command

bloodearnest:
This will take a path to a project.yaml file, and use that to inform the
tool as to which images should be pulled locally.

The primary motivation for this is faster setup of development
environments, particularly with gitpod.io in mind.

### DIFF
--- a/opensafely/pull.py
+++ b/opensafely/pull.py
@@ -32,18 +32,18 @@ def add_arguments(parser):
         help="Update docker images even if not present locally",
     )
     parser.add_argument(
-        "--preseed",
+        "--project",
         help="Use this project to yaml to decide which images to download",
     )
 
 
-def main(image="all", force=False, preseed=None):
+def main(image="all", force=False, project=None):
     if not docker_preflight_check():
         return False
 
-    if preseed:
+    if project:
         force = True
-        images = get_actions_from_project_file(preseed)
+        images = get_actions_from_project_file(project)
     elif image == "all":
         images = IMAGES
     else:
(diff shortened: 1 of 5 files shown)


opensafely/pull.py
```diff
@@ -28,13 +31,20 @@ def add_arguments(parser):
         action="store_true",
         help="Update docker images even if not present locally",
     )
+    parser.add_argument(
+        "--preseed",
```
ghickman, line 35: I don't know what the correct terms are here but I'm pretty sure `pre` (it's a prefix but is that the right term when talking about language?) should be hyphenated (eg `--pre-seed`), not gating the review on it though.

opensafely/pull.py
```diff
@@ -62,6 +72,35 @@ def main(image, force):
...
+            continue
+
+        name, _, version = command.partition(":")
+        if name in IMAGES:
+            images.append(name)
+
+    if not images:
+        raise RuntimeError(f"No actions found in {project_yaml} to preseed")
+
+    return images
```
ghickman, line 101: I'm reminded that we really need to build a project.yaml parsing library!  Not something for this PR though.

tests/test_pull.py
```diff
@@ -94,13 +119,29 @@ def test_remove_deprecated_images(run):
     pull.remove_deprecated_images(local_images)
 
 
+@pytest.mark.parametrize("project_yaml,exc_msg", [
+    ("doesnotexist.yaml", "Could not find"),
+    ("bad.yaml", "Could not parse"),
+    ("noactions.yaml", "No actions found"),
+])
```
ghickman, line 126: This is really nice 👍

ghickman (approved):
Couple of comments and a suggestion on the switch, but I'll leave that up to you to make the call on 👍

opensafely/pull.py
bloodearnest: Hmm, it's commonly `preseed` in debian/ubuntu world, e.g. https://help.ubuntu.com/lts/installation-guide/s390x/apb.html

I wonder if `--project` would be better anyway?

opensafely/pull.py
bloodearnest: I'm reminded of that frequently!

tests/test_pull.py
bloodearnest: Thanks. I'm a big fan of ~2 pytest tests for a function. 1 to test the happy path, and a parameterised one that tests the error paths, which should all raise an exception of some kind. I like partial exc message matching too, as it reduces test brittleness.

opensafely/pull.py
bloodearnest: Have changed to `--project`.

tests/test_pull.py
ghickman: /me side eyes the many failure tests for job-sever

🤔

bloodearnest: Sometimes its not always doable, but I've found it a good default for smaller functions.

For more complex cases, I use a generator function to generate the test cases.

PR #52, opened 2021-06-17
ci: update branch name in workflow

StevenMaude:
Before we switch the default branch name.


StevenMaude:
Requested as per https://github.com/opensafely-core/project-tidy-up/issues/30

I don't have permissions to rename the default branch from `master` to `main`, so a kindly repository owner should do that.

evansd (approved):

PR #66, opened 2021-09-30
Update job-runner version

rebkwok:


benbc (commented):
Shouldn't this come with a complete `jobrunner` update and a change in `vendor.txt`?

rebkwok:
> Shouldn't this come with a complete `jobrunner` update and a change in `vendor.txt`?

Duh, yes, you're right, it should.  I'd forgotten that was where the change was.

evansd (approved):

PR #68, opened 2021-10-21
jobrunner 2.36.0

bloodearnest:
Updated vendored jobrunner to 2.36.0

This change pulls in the new executor API, but it is not enabled by default. It has been tested and seems to work fine on a the couple of projects I tried.

This change also pulls in the migration of local workspace, but should also be fine, as has worked against every issue we found in prod.


bloodearnest:
Done

PR #70, opened 2021-11-11
test: Test with Python 3.10

StevenMaude:
I was looking at https://github.com/opensafely/documentation/issues/485 and wondering what Python version someone might get if they install Python from Microsoft Store.

It's possible (though can't check) that soon-ish they'd get Python 3.10 when searching for Python.

So, it seems wise to run the tests with Python 3.10 here, because:

* the tests work (with a couple of dependency upgrades here)
* to flag any potential future issues that we might miss otherwise


evansd (approved):
Looks good, thanks!

PR #71, opened 2021-11-11
chore: Update URLs in setup.py

StevenMaude:
GitHub redirects the previous URL. But it's still nice to be directly
correct.


ghickman (approved):
I wonder if we should provide some sidebar links via setup.py too?  It's possible to add links directly to documentation, etc this way.

StevenMaude:
Relevant on what project URLs give sidebar icons: https://github.com/simonw/til/blob/51147f9726f717109fc7d9046f8f7cb1e424f66a/pypi/project-links.md

StevenMaude:
@ghickman: how does that look? Or did you have other URLs in mind?

ghickman (approved):
👌

PR #75, opened 2021-11-19
Check-datasource

Jongmassey:
Implementation of ideas originally expressed in [bash script](https://github.com/Jongmassey/research_repository_permissions) to allow for checking study definitions for usage of restricted datasets

### DIFF
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -41,15 +41,15 @@ def main():
 
 
 def print_violations(found_datasets):
-    s =("Usage of restricted datasets found:\n")
+    s = "Usage of restricted datasets found:\n"
     for d, functions in found_datasets.items():
-        s= s+(f"{d}:\n")
+        s = s + (f"{d}:\n")
         for fn, files in functions.items():
-            s= s+(f"- {fn}\n")
+            s = s + (f"- {fn}\n")
             for f, lines in files.items():
-                s=s+(f"  - {f}:\n")
+                s = s + (f"  - {f}:\n")
                 for ln, line in lines.items():
-                    s=s+(f"    line {ln}: {line}\n")
+                    s = s + (f"    line {ln}: {line}\n")
     return s
 
 
@@ -77,7 +77,7 @@ def check_file(filename, regex):
 
 
 def get_local_py_files():
-    return glob.glob("**/*.py",recursive=True)
+    return glob.glob("**/*.py", recursive=True)
 
 
 def get_datasource_permissions(permissions_url):
@@ -97,13 +97,13 @@ def get_repository_name():
     else:
 
         url = subprocess.run(
-            args=["git", "config", "--get","remote.origin.url"],
+            args=["git", "config", "--get", "remote.origin.url"],
             capture_output=True,
             text=True,
         ).stdout
         return (
             url.replace("https://github.com/", "")
-            .replace("git@github.com:","")
+            .replace("git@github.com:", "")
             .replace(".git", "")
             .strip()
         )
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -1,9 +1,114 @@
+from opensafely._vendor import requests
+from opensafely._vendor.ruamel.yaml import YAML
+from os import environ, popen, walk, path
+import re
+
+
 DESCRIPTION = "Check the opensafely project for correctness"
+
+RESTRICTED_DATASETS = {"ICNARC": ["admitted_to_icu"]}
+
+# todo: dummy location and data
+PERMISSIONS_URL = "https://raw.githubusercontent.com/Jongmassey/research_repository_permissions/main/repository_permissions.yaml"
 
 
 def add_arguments(parser):
-    ...
+    pass
 
 
 def main():
-    ...
+    repo_name = get_repository_name()
+    permissions = get_datasource_permissions(PERMISSIONS_URL)
+    allowed_datasets = get_allowed_datasets(repo_name, permissions)
+    datasets_to_check = {
+        k: v
+        for k, v in RESTRICTED_DATASETS.items()
+        if k not in allowed_datasets
+    }
+    files_to_check = get_local_py_files()
+
+    found_datasets = {
+        dataset: check_dataset(functions, files_to_check)
+        for dataset, functions in datasets_to_check.items()
+    }
+
+    if found_datasets:
+        print_violations(found_datasets)
+        exit(1)
+    else:
+        print("Success")
+        exit(0)
+
+
+def print_violations(found_datasets):
+    print('Usage of restricted datasets found:')
+    for d,files in found_datasets.items():
+        print(f'{d}:')
+        for f,lines in files.items():
+            print(f'\t {f} :')
+            for ln, line in lines:
+                print(f"\t\t line {ln}: {line}")
+
+
+def check_dataset(functions, files_to_check):
+    found_functions = {}
+    for function in functions:
+        regex = re.compile(f"\.{function}\(")
+        found_files = {}
+        for f in files_to_check:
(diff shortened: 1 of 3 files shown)


Jongmassey:
todo: deploy repository permissions yaml somewhere proper and update URL in check.py

Jongmassey:
tests are failing in action because github's environment variables are set, fix inbound

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+        if check_dataset(functions, files_to_check)
+    }
+
+    if found_datasets:
+        exit(print_violations(found_datasets))
+    else:
+        print("Success")
+
+
+def print_violations(found_datasets):
```
bloodearnest, line 43: nit: this doesn't actually print anything any more, so maybe `format_violations`?

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+    s = "Usage of restricted datasets found:\n"
+    for d, functions in found_datasets.items():
+        s = s + (f"{d}:\n")
+        for fn, files in functions.items():
+            s = s + (f"- {fn}\n")
+            for f, lines in files.items():
+                s = s + (f"  - {f}:\n")
+                for ln, line in lines.items():
+                    s = s + (f"    line {ln}: {line}\n")
```
bloodearnest, line 52: nit: strings are immutable in python, so string concatenation is generally avoided as it allocation heavy.

The typical idiom is to join a list of lines, e.g.

```python
lines = []
...
lines.append(line)
...
return "\n".join(lines)
```

Which also saves you have having to append "\n" to every line's text.

However, for more fun, you can stream it!

```python
def print_violations(found_datasets):
    yield "Usage of restricted datasets found:"
    for d, functions in found_datasets.items():
        yield f"{d}:"
        for fn, files in functions.items():
            yield f"- {fn}"
            for f, lines in files.items():
                yield f"  - {f}:"
                for ln, line in lines.items():
                    yield f"    line {ln}: {line}"

```

And use it like so:

```python
voilations = "\n".join(print_violations(found_violations))
sys.exit(violations)
```

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+    files_to_check = get_local_py_files()
+
+    found_datasets = {
+        dataset: check_dataset(functions, files_to_check)
+        for dataset, functions in datasets_to_check.items()
+        if check_dataset(functions, files_to_check)
+    }
+
+    if found_datasets:
+        exit(print_violations(found_datasets))
```
bloodearnest, line 38: nit: python convention would fully qualify the exit.

It is a global function too, but that's really intended for interactive console use.

```python
import sys
...
sys.exit(...)
```

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+        if found_files:
+            found_functions[function] = found_files
+    return found_functions
+
+
+def check_file(filename, regex):
+    found_lines = {}
+    with open(filename, "r") as f:
+        for ln, line in enumerate(f.readlines(), start=1):
+            if regex.search(line) and not line.lstrip().startswith("#"):
```
bloodearnest, line 74: nit: I would write this like so:

```python
if line.lstrip().startswith("#"):
    continue
if regex.search(line):
    ...
```
My thinking is that a) logicially, its just skipping comment lines, which feels right to me b) it avoids running the regex on comment lines.

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+                found_lines[ln] = line
+    return found_lines
+
+
+def get_local_py_files():
+    return glob.glob("**/*.py", recursive=True)
```
bloodearnest, line 80: nit: is this one line worth a function? I'd probably just call it directly.

tests/test_check.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+STUDY_DEF_HEADER = """from cohortextractor import (StudyDefinition, patients)
+study = StudyDefinition ("""
+
+
+def write_restricted_files(path):
+    for a in [1, 2]:
+        with open(
+            os.path.join(str(path), f"study_definition_restricted_{a}.py"), "w"
```
bloodearnest, line 63: tmp_path is an instance of pathlib.Path, and is the newer preferred way to interact with the file system.
So rather than converting it to a string and using os.path.*, I'd do this:

```python
    ....
    f = path / f"study_definition_restriction_{a}.py"
    f.write_text(textwrap.dedent(f"""\
        {STUDY_DEF_HEADER}
        a={RESTRICTED_FUNCTION_CALL},
        #b={RESTRICTED_FUNCTION_CALL},
        c={UNRESTRICTED_FUNCTION_CALL},
        #d={UNRESTRICTED_FUNCTION_CALL},
    """)
    ...
```

tests/test_check.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+def test_unrestricted_local_norepo(tmp_path, capsys, monkeypatch):
+    if "GITHUB_REPOSITORY" in os.environ:
+        monkeypatch.delenv("GITHUB_REPOSITORY")
+    prev_dir = os.getcwd()
+    os.chdir(tmp_path)
+    write_unrestricted_files(tmp_path)
+    validate_pass(capsys)
+    os.chdir(prev_dir)
```
bloodearnest, line 123: ok, so this undoing won't work in the case of test failure.

I suggest a fixture:

```python
@pytest.fixture
def repo_dir(tmp_path):
    old = os.getcwd()
    os.chdir(tmp_path)
    yield tmp_path
    os.chdir(old)
```

And then use that instead of `tmp_path` fixture. That way, any test with `repo_dir` fixture will have the correct cleanup behaviour, and also you can use `repo_dir` exactly like `tmp_path`.

opensafely/check.py
```diff
@@ -0,0 +1,115 @@
...
+import re
+import glob
+import subprocess
+
+DESCRIPTION = "Check the opensafely project for correctness"
+
+RESTRICTED_DATASETS = {"icnarc": ["admitted_to_icu"]}
+
+# todo: dummy location and data
+PERMISSIONS_URL = "https://raw.githubusercontent.com/Jongmassey/research_repository_permissions/main/repository_permissions.yaml"
```
bloodearnest, line 13: I think we need to find a proper home for this file before we land it.

We could also supply it via environment variable with a default value, which can be useful for testing.

bloodearnest (commented):
Looking good.

Left a few of comments on things I think we should fix, mainly the proper location of the rules file, and a couple of tests fixes (the `repo_dir` fixture and using the features of pathlib.Path in tests).

I've also left a bunch of 'nit:' comments. These are not blockers, I'd be happy to land without changing them, but thought you might appreciate pointers on some more idomatic pythonic ways of doing things, but it's up to you!

opensafely/check.py
Jongmassey: It didn't used to be but became one when refactored. As it's only used in one place I'll just put it in directly.

opensafely/check.py
Jongmassey: Good to know a concise way of streaming it in Python. I miss my StringBuider classes!

Jongmassey:
>the proper location of the rules file
https://github.com/opensafely-core feels like the natural organisation for it. It doesn't really belong in this repository, or rather it might be harder for non-developer users to update permissions here.

opensafely/check.py
Jongmassey: https://github.com/opensafely-core feels like the natural organisation for it. It doesn't really belong in this repository, or rather it might be harder for non-developer users to update permissions here.

bloodearnest (approved):

PR #79, opened 2021-11-25
Run 'check' in warn mode with 'run'

Jongmassey:


opensafely/check.py
```diff
@@ -14,10 +14,14 @@
 
 
 def add_arguments(parser):
-    pass
+    parser.add_argument(
+        "--error",
+        help="Error (exit 1) if restricted datasets found",
+        action="store_true",
```
bloodearnest, line 20: What's the thinking behind exposing this flag? 

I think the error behaviour should probably be the default for the `check` command, so if anything it would be `--no-error`. In which case, I'm not sure it's worth adding. We don't need to turn if off if we invoke it via `opensafely check`.

opensafely/check.py
```diff
@@ -14,10 +14,14 @@
...
+    )
 
 
-def main():
+def main(error=False):
```
bloodearnest, line 24: I'd be tempted to make the default error=True, and only disable it in the special case of local `run`

bloodearnest (approved):

opensafely/check.py
Jongmassey: That seems sensible as there's literally only one use case where we want it to continue rather than error

opensafely/check.py
Jongmassey: Agree

Issue #80, opened 2021-12-03
Add preflight check for the presence of `git`


evansd:
We shell out to `git` in multiple places, so we should check that it's accessible beforehand so we can error out with a helpful message (as we currently do for docker).

This is particularly important because although Github Desktop bundles the `git` binary, it doesn't put it on the path so installing Github Desktop is insufficient here. We could try to spot the Github Desktop is installed and get the path to `git` ourselves but I think that's probably a bad idea. Better to make it clear in the error message that installing Github Desktop is insufficient and link to the download page for vanilla git:
https://git-scm.com/downloads

Bonus points for detecting the operating system and linking to the correct sub-page.

Associated Slack thread where this issue first came up:
https://ebmdatalab.slack.com/archives/C010SJ89SA3/p1638537319387800

evansd:
For reference, this is why Github Desktop doesn't put their bundled binary on the path:
https://github.com/desktop/desktop/issues/3708#issuecomment-354665183

Jongmassey:
Documentation updated to reflect need for `git` in addition to GitHub Desktop

bloodearnest:
If we do detect it is missing, should point to documentation on how to install it, especially on windows, as there a nuances, as the above GH desktop link captures.

PR #81, opened 2021-12-07
fix: gracefully fail when git not installed

Jongmassey:
users may not have git installed, or available on path. In this case, don't throw error but instead proceed as if repo name is unavailable.

### DIFF
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -100,18 +100,20 @@ def get_repository_name():
     if "GITHUB_REPOSITORY" in os.environ:
         return os.environ["GITHUB_REPOSITORY"]
     else:
-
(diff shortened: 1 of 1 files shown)


evansd:
Ah, thanks @Jongmassey. I'm not sure though that this is quite the right approach because this isn't the only place we shell out to git: we also use it for pulling in the OpenSAFELY Stata license (where applicable) and for running reusable actions. And it's quite possible we'll add other uses in future. So I think we really do need to require `git` and therefore we'll need something like the preflight check described here:
 * #80

Jongmassey:
OK, I'll pass this feedback to the user who encountered the issue. Thanks.

PR #84, opened 2022-01-10
feat: add an initial jupyter command

bloodearnest:
This will run jupyter lab in the current directory using the opensafely
python image. It should behave the same as running `jupyter lab`
manually.

### DIFF
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -1,10 +1,9 @@
 import argparse
-from pathlib import Path
 import sys
+from pathlib import Path
 
+from opensafely import check, codelists, jupyter, pull, upgrade
 from opensafely._vendor.jobrunner.cli import local_run
-from opensafely import codelists, pull, upgrade, check
-
 
 __version__ = Path(__file__).parent.joinpath("VERSION").read_text().strip()
 
@@ -25,9 +24,7 @@ def main():
         title="available commands", description="", metavar="COMMAND"
     )
 
-    parser_help = subparsers.add_parser(
-        "help", help="Show this help message and exit"
-    )
+    parser_help = subparsers.add_parser("help", help="Show this help message and exit")
     parser_help.set_defaults(function=show_help)
 
     # Add `run` subcommand
@@ -36,9 +33,7 @@ def main():
     local_run.add_arguments(parser_run)
 
     # Add `codelists` subcommand
-    parser_codelists = subparsers.add_parser(
-        "codelists", help=codelists.DESCRIPTION
-    )
+    parser_codelists = subparsers.add_parser("codelists", help=codelists.DESCRIPTION)
     parser_codelists.set_defaults(function=codelists.main)
     codelists.add_arguments(parser_codelists)
 
@@ -56,6 +51,11 @@ def main():
     parser_check = subparsers.add_parser("check", help=check.DESCRIPTION)
     parser_check.set_defaults(function=check.main)
     check.add_arguments(parser_check)
+
+    # Add `jupyter` subcommand
+    parser_jupyter = subparsers.add_parser("jupyter", help=jupyter.DESCRIPTION)
+    parser_jupyter.set_defaults(function=jupyter.main)
+    jupyter.add_arguments(parser_jupyter)
 
     # we version check before parse_args is called so that if a user is
     # following recent documentation but has an old opensafely installed,
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -10,7 +10,6 @@ __version__ = Path(__file__).parent.joinpath("VERSION").read_text().strip()
 
 def main():
     parser = argparse.ArgumentParser()
-    supports_unknown_args = set()
 
     def show_help(**kwargs):
         parser.print_help()
@@ -29,11 +28,12 @@ def main():
     parser_help.set_defaults(function=show_help)
 
     def add_subcommand(cmd, module):
-        subparser = subparsers.add_parser(cmd, help=local_run.DESCRIPTION)
-        subparser.set_defaults(function=module.main)
-        handles_unknown = module.add_arguments(subparser)
-        if handles_unknown:
-            supports_unknown_args.add(module.main)
+        subparser = subparsers.add_parser(cmd, help=module.DESCRIPTION)
+        subparser.set_defaults(
+            handles_unknown_args=False, 
+            function=module.main,
+        )
+        module.add_arguments(subparser)
 
     add_subcommand("run", local_run)
     add_subcommand("codelists", codelists)
@@ -52,8 +52,7 @@ def main():
     # start by using looser parsing only known args, so we can get the sub command
     args, unknown = parser.parse_known_args()
 
-    #
-    if args.function in supports_unknown_args:
+    if args.handles_unknown_args:
         kwargs = vars(args)
         kwargs["unknown_args"] = unknown
     # allow supparsers to opt out of looser parser_known_args parsing
@@ -63,6 +62,7 @@ def main():
         kwargs = vars(args)
 
     function = kwargs.pop("function")
+    kwargs.pop("handles_unknown_args")
     success = function(**kwargs)
 
     # if `run`ning locally, run `check` in warn mode
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -10,6 +10,7 @@ __version__ = Path(__file__).parent.joinpath("VERSION").read_text().strip()
 
 def main():
     parser = argparse.ArgumentParser()
+    supports_unknown_args = set()
 
     def show_help(**kwargs):
         parser.print_help()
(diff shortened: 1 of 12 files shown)


opensafely/__init__.py
```diff
@@ -57,6 +57,11 @@ def show_help(**kwargs):
     parser_check.set_defaults(function=check.main)
     check.add_arguments(parser_check)
 
+    # Add `notebook` subcommand
+    notebook_check = subparsers.add_parser("notebook", help=notebook.DESCRIPTION)
+    notebook_check.set_defaults(function=notebook.main)
```
evansd, line 62: I think you want `parser_notebook` here if you're going for consistency with the existing code.

opensafely/notebook.py
```diff
@@ -0,0 +1,71 @@
...
+            "0.0.0.0",
+            # python image currently runs as root
+            "--allow-root",
+            # disable authentication, as it's only accessible via direct access
+            # to the docker container
+            "--LabApp.token=''",
+        ]
+
+    launch_ps = docker(*launch)
+    container = launch_ps.stdout.strip()
```
evansd, line 54: I don't actually think you need `container` because you can pass `name` to `docker inspect` and it will work exactly the same.

opensafely/notebook.py
```diff
@@ -0,0 +1,71 @@
...
+    # let jupyter get started
+    time.sleep(1)
+
+    # figure out the url
+    inspect_ps = docker("inspect", container)
+    metadata = json.loads(inspect_ps.stdout)[0]
+    ip = metadata["NetworkSettings"]["IPAddress"]
+    url = f"http://{ip}:8888"
```
iaindillingham, line 62: Maybe one for later, but @HelenCEBM mentioned that it would be useful to have multiple instances of JupyterLab open at the same time, but for different studies (and so different bind directories).

iaindillingham (approved):
🚀

opensafely/notebook.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # open a webbrowser pointing to the docker container
+    webbrowser.open(url, new=2)
+
+    # show what the server has said so far
+    docker("logs", container, capture_output=False)
```
evansd, line 68: Given you don't need to read stdout to get `container`, do you need to capture output from the initial command at all? That would also avoid the big silent wait you'll get if you run this without having the `python` image downloaded already.

opensafely/notebook.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # take over the shell so we see further info from the jupyter server
+    docker("container", "attach", container, capture_output=False)
```
evansd, line 71: I did [something](https://github.com/opensafely/server-instructions/blob/master/scripts/database-notebook.sh) sort of similar to this for running notebooks on the TPP server with database access. And I seem to remember that shutting down Jupyter via the web interface still leaves the container running on Windows, although it terminates it on Linux.

It might be worth waiting (or just looping on sleep) for the process in Python and then explicitly terminating the container on exit.

evansd (approved):
:+1: Such a great bit of itch-scratching! A few thoughts/comments below

opensafely/notebook.py
bloodearnest: This works.

The ip address here is the docker container IP, so is unique to each jupyter instance. I deliberately didn't map the ports for this reason.

opensafely/__init__.py
bloodearnest: oh, c/p error yes

opensafely/notebook.py
bloodearnest: oh! yes, brilliant. I only added name at the end.

opensafely/notebook.py
bloodearnest: Huh, that's an interesting idea. I think it should work.

opensafely/notebook.py
bloodearnest: Ah, I haven't tried that, will try this whole thing on windows

opensafely/notebook.py
bloodearnest: So, this is trickier than I thought.

We are using --detach to also fork. If we want to do that in python, and preserve stdin/out, we need to use subprocess.Popen, as .run always waits till completion.

I had a quick try with Popen, but it gets ugly quickly, managing two programs writing to stdout  gets a very messy, tracebackes with weird formatting and doing funny things we stdint, forcing me to run `reset` in my terminal

bloodearnest: We could do this in a thread, ofc...

evansd: Ah yes, I can see that's going to get tricky. I guess one way of avoiding the silent wait if the image needs to download would be to split the `run` into separate `create` and `start` calls. The `create` can be blocking and non-capturing so you get the download progress stuff. And then the `start` can be detached. Maybe all getting a bit complicated though.

If you did go down the threading route then you spin off a thread which just polls waiting for named container to come up, then gets its IP and opens a browser. And then the main thread just needs a single call to (non-detached) `docker run`.

bloodearnest: Great minds and all that, I'd already started down that thread route :)

opensafely/notebook.py
bloodearnest: In linux, it does shutdown. About to try in on windows.

opensafely/__init__.py
```diff
@@ -27,35 +28,19 @@ def show_help(**kwargs):
     parser_help = subparsers.add_parser("help", help="Show this help message and exit")
     parser_help.set_defaults(function=show_help)
 
-    # Add `run` subcommand
-    parser_run = subparsers.add_parser("run", help=local_run.DESCRIPTION)
-    parser_run.set_defaults(function=local_run.main)
-    local_run.add_arguments(parser_run)
+    def add_subcommand(cmd, module):
+        subparser = subparsers.add_parser(cmd, help=local_run.DESCRIPTION)
```
evansd, line 32: This should be `module.DESCRIPTION`

opensafely/__init__.py
```diff
@@ -27,35 +28,19 @@ def show_help(**kwargs):
...
-    # Add `jupyter` subcommand
-    parser_jupyter = subparsers.add_parser("jupyter", help=jupyter.DESCRIPTION)
-    parser_jupyter.set_defaults(function=jupyter.main)
-    jupyter.add_arguments(parser_jupyter)
+    add_subcommand("run", local_run)
+    add_subcommand("codelists", codelists)
+    add_subcommand("pull", pull)
+    add_subcommand("upgrade", upgrade)
+    add_subcommand("check", check)
+    add_subcommand("jupyter", jupyter)
```
evansd, line 43: +1 for factoring this stuff out into a function, it's so obviously better this way!

opensafely/__init__.py
```diff
@@ -27,35 +28,19 @@ def show_help(**kwargs):
...
-    # Add `run` subcommand
-    parser_run = subparsers.add_parser("run", help=local_run.DESCRIPTION)
-    parser_run.set_defaults(function=local_run.main)
-    local_run.add_arguments(parser_run)
+    def add_subcommand(cmd, module):
+        subparser = subparsers.add_parser(cmd, help=local_run.DESCRIPTION)
+        subparser.set_defaults(function=module.main)
+        handles_unknown = module.add_arguments(subparser)
+        if handles_unknown:
+            supports_unknown_args.add(module.main)
```
evansd, line 36: I think there might be a neater way of doing this which is to call:
```py
subparser.set_defaults(handles_unknown_args=False)
module.add_arguments(subparser)
```

And then if the module wants to opt in to looser parsing its `add_arguments` function can call:
```py
subparser.set_defaults(handles_unknown_args=True)
```

Then there's no need to track this in a set, and no need to have the boolean return value from `add_arguments`.

opensafely/jupyter.py
```diff
@@ -0,0 +1,104 @@
...
+
+def open_browser(name, port):
+
+    try:
+        # wait for container to be up
+        while True:
+            ps = subprocess.run(
+                ["docker", "inspect", name],
+                capture_output=True,
+                universal_newlines=True,
```
evansd, line 40: Just a by-the-way, but `text=True` is the new way of spelling this as `universal_newlines` was considered not readable enough (kids these day, etc).

opensafely/jupyter.py
```diff
@@ -0,0 +1,104 @@
...
+        "--init",
+        "-it",
+        f"--name={name}",
+        f"--hostname={name}",
+        "--label=opensafely",
+        f"-v=/{pwd}:/workspace",
+        "ghcr.io/opensafely-core/python",
+        # we wrap the jupyter command in a bash invocation, so we can use
+        # hostname -i to find out the containers IP address, which allows us to
+        # set the correct url to show in the output.
```
evansd, line 97: :ok_hand:

evansd (approved):
Looks great. Just one typo that needs fixing (the `module.HELP` thing) and a possible small (optional) refactoring. It's amazing that adding threads has made this simpler and more understandable than before.

opensafely/__init__.py
bloodearnest: whoo boy, thanks

opensafely/__init__.py
bloodearnest: Ah nice, will do

opensafely/jupyter.py
bloodearnest: Yeah, I suspect some IDE tools expect to parse the output of jupyter to find where it's running

opensafely/jupyter.py
```diff
@@ -0,0 +1,181 @@
...
+import time
+import webbrowser
+from pathlib import Path
+from urllib import request
+
+DESCRIPTION = "Run a jupyter lab notebook using the OpenSAFELY environment"
+
+
+# quick and directy debugging hack, as due to threads and windos this is tricky
+# to debug
```
evansd, line 18: This comment seems a bit free-floating

opensafely/jupyter.py
```diff
@@ -0,0 +1,181 @@
...
+        "--rm",
+        "--init",
+        "-it",
+        # we use our port on both sides of the docker port mapping so that
+        # jupyter has all the info
+        f"-p={port}:{port}",
+        f"--name={name}",
+        f"--hostname={name}",
+        "--label=opensafely",
+        # note: on windows this will preserve drive letter, but swtitch to unix
```
evansd, line 174: typo: `swtitch`

evansd (approved):
Looks good, if inevitably more complicated than we all thought it would be! There are a few bits that could be tidied up but nothing blocking

PR #91, opened 2022-01-19
fix: add more debug output to jupyter command

bloodearnest:
It's hard to test this on different users setups, so make more debugging
output available.

### DIFF
--- a/opensafely/jupyter.py
+++ b/opensafely/jupyter.py
@@ -13,6 +13,18 @@ from urllib import request
 DESCRIPTION = "Run a jupyter lab notebook using the OpenSAFELY environment"
 
 
+# poor mans debugging because debugging threads on windows is hard
+if os.environ.get("DEBUG", False):
+    def debug(msg):
+        # threaded output for some reason needs the carriage return or else
+        # it doesn't reset the cursor.
+        sys.stderr.write("DEBUG: " + msg.replace("\n", "\r\n") + "\r\n")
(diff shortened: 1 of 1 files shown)


opensafely/jupyter.py
```diff
@@ -13,6 +13,18 @@
...
+# poor mans debugging because debugging threads on windows is hard
+if os.environ.get("DEBUG", False):
+    def debug(msg):
+        # threaded output for some reason needs the carriage return or else
+        # it doesn't reset the cursor.
+        sys.stderr.write("DEBUG: " + msg.replace("\n", "\r\n") + "\r\n")
+        sys.stderr.flush()
+else:
+    def debug(msg):
+        pass
```
ghickman, line 25: Given the function name and parameters are the same for both forks, is it worth pushing the if-block inside the function?

Something like:

```
def debug(msg):
    DEBUG = os.environ.get("DEBUG", False)
    if not DEBUG:
        return

    # true branch
    …
```

ghickman (approved):
One small thought on layout of the `debug` function, feel free to ignore!

opensafely/jupyter.py
bloodearnest: Why do 20 `if` s if you can do one!

bloodearnest: Also, I plan to DTRT thing an add debug logging and -vvv switches at somepoint, so this is only temporary

PR #93, opened 2022-01-19
fix: jupyter another attempt at dealing with tty in git-bash

bloodearnest:
The goal is to automatically wrap our call to "docker run -it ..." in
winpty if we detect we are running in git-bash's broken default
terminal.

Both cmd.exe and Windows Terminal provide a proper tty so do not need
this.

The problem is that passing the `-v=...` paths as arguments to winpty
forces them through git-bash path translation, which we had previously
been able to avoid as we were going via subprocess.run(), and the path
translation mucked up the -v mappings.

So, this change does two thinks:

1) try to detect when were are in the severely limit git-bash/mingw
   environement, and wrap the docker command in winpty only then.

2) Escape the docker side of the mount from git-bash with
   `://workspace`, and revert to using standard backslash paths on the
   host side.

This has been manually tested in git-bash and in cmd.exe and worked on
a least one windows machine.

### DIFF
--- a/opensafely/jupyter.py
+++ b/opensafely/jupyter.py
@@ -1,5 +1,6 @@
 import json
 import os
+import platform
 import shutil
 import socket
 import subprocess
@@ -15,12 +16,16 @@ DESCRIPTION = "Run a jupyter lab notebook using the OpenSAFELY environment"
 
 # poor mans debugging because debugging threads on windows is hard
 if os.environ.get("DEBUG", False):
+
     def debug(msg):
         # threaded output for some reason needs the carriage return or else
         # it doesn't reset the cursor.
         sys.stderr.write("DEBUG: " + msg.replace("\n", "\r\n") + "\r\n")
         sys.stderr.flush()
+
+
 else:
+
     def debug(msg):
         pass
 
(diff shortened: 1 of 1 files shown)


opensafely/jupyter.py
```diff
@@ -15,12 +16,16 @@
...
 # poor mans debugging because debugging threads on windows is hard
 if os.environ.get("DEBUG", False):
+
     def debug(msg):
         # threaded output for some reason needs the carriage return or else
         # it doesn't reset the cursor.
         sys.stderr.write("DEBUG: " + msg.replace("\n", "\r\n") + "\r\n")
         sys.stderr.flush()
+
+
```
StevenMaude, line 26: nit: I probably wouldn't have this whitespace here, but that's a personal style choice. If you prefer it, that's also fine.

opensafely/jupyter.py
```diff
@@ -163,19 +169,28 @@ def main(directory, name, port, no_browser, unknown_args):
...
         "ghcr.io/opensafely-core/python",
     ]
 
-    # on windows, we need to wrap command in winpty to for docker run -it
-    winpty = shutil.which("winpty")
-    if winpty is not None:
-        debug(f"found winpty at {winpty}")
+    # git-bash's default terminal does not support a unix tty directly. So, try
+    # and detect this situation and wrap it ourselves
+    tty = (
```
StevenMaude, line 179: nit: I'd probably order these assignments as in the `if` below. The logical order does feel like should be in that `if`, that is:

* is this Windows?
* is winpty available?
* is this running in a tty?

You could also maybe skip the other checks entirely if Windows isn't running. But these look very inexpensive, so not a big problem either way.

StevenMaude (approved):

opensafely/jupyter.py
bloodearnest: Huh, yes, odd choice on my part

bloodearnest: It was black!

opensafely/jupyter.py
bloodearnest: I've refactored into a function to make this cleaner.

opensafely/jupyter.py
bloodearnest: I'm innocent! :)

PR #94, opened 2022-01-19
fix: read git config instead of using git cli

Jongmassey:
Github desktop is the recommended installation of git as per the opensafely docs.
It does not add `git` to the Windows path. Therefore when `opensafely check` tries to call git to ascertain the repository name, this can fail.

This PR parses the git config file instead.

### DIFF
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -23,6 +23,8 @@ def main(continue_on_error=False):
         os.environ.get("OPENSAFELY_PERMISSIONS_URL") or PERMISSIONS_URL
     )
     repo_name = get_repository_name()
+    if not repo_name and not continue_on_error:
+        sys.exit("Unable to find repository name")
     permissions = get_datasource_permissions(permissions_url)
     allowed_datasets = get_allowed_datasets(repo_name, permissions)
     datasets_to_check = {
@@ -103,10 +105,16 @@ def get_repository_name():
     else:
         git_config_path = Path(".git", "config")
         if not git_config_path.is_file():
+            print("Git config file not found")
             return
         config = configparser.ConfigParser()
-        config.read(git_config_path)
+        try:
+            config.read(git_config_path)
+        except Exception as e:
+            print(f"Unable to read git config.\n{str(e)}")
+            return
         if 'remote "origin"' not in config.sections():
+            print("Remote 'origin' not defined in git config.")
             return
         url = config['remote "origin"']["url"]
         return (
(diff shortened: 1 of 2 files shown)


opensafely/check.py
```diff
@@ -97,12 +101,14 @@ def get_repository_name():
...
-        url = subprocess.run(
-            args=["git", "config", "--get", "remote.origin.url"],
-            capture_output=True,
-            text=True,
-        ).stdout
+        git_config_path = Path(".git", "config")
+        if not git_config_path.is_file():
+            return
+        config = configparser.ConfigParser()
+        config.read(git_config_path)
```
StevenMaude, line 108: thought 1: Let's say the file exists, but `config.read()` fails for some obscure reason (maybe the config file is locked by some other process). Then you'll simply get no remote in the same way as you would if the file didn't exist.

There are several different failure modes:

* no `.git/config`
* can't read `.git/config`
* no `remote "origin"` in `.git/config`

Maybe it's useful to distinguish these to the user, but it may well not be.

opensafely/check.py
```diff
@@ -97,12 +101,14 @@ def get_repository_name():
...
+        if 'remote "origin"' not in config.sections():
+            return
+        url = config['remote "origin"']["url"]
```
StevenMaude, line 111: thought 2: do we know for sure that the remote will be "origin" for users? I can arbitrarily choose a name for the remote. If we're reasonably confident, that's fine!

This is another reason why you might want to notify the user what failed to be found and why.

StevenMaude (approved):
I added a couple of thoughts about the approach as a whole, and making failures explicit to the user. But those are from a naive position. Feel free to pass them by, if they're unhelpful.

Overall, this looks fine as a direct change for the existing code.

opensafely/check.py
bloodearnest: It's our only option, I think.

bloodearnest:
It might be worth handling the errors for the opportunistic drive-by call to check() that's embedded in the `run` command vs the explicit `opensafely check` command. The former should never cause the the run to fail, where as the latter should exit with a failure.

opensafely/check.py
```diff
@@ -103,10 +105,16 @@ def get_repository_name():
...
         git_config_path = Path(".git", "config")
         if not git_config_path.is_file():
+            print("Git config file not found")
             return
         config = configparser.ConfigParser()
-        config.read(git_config_path)
+        try:
+            config.read(git_config_path)
+        except Exception as e:
+            print(f"Unable to read git config.\n{str(e)}")
```
bloodearnest, line 114: I'm not sure we should print this level of detail for the drive by check, so maybe only print these if continue_on_error is False?

PR #98, opened 2022-02-03
Restrict therapeutics dataset

### LINKED ISSUE #97: Add antiviral eligibility functions to restricted list
Once https://github.com/opensafely-core/cohort-extractor/issues/713 is in progress and function names decided, these should be added to https://github.com/opensafely-core/opensafely-cli/blob/main/opensafely/check.py#L12 before deployment

rebkwok:
Fixes #97 

Adds the therapeutics dataset and its associated function (`with_covid_therapeutics`) to the restricted datasets list.

As a side effect, the tests now mock the output of the call to the test permissions URL (it calls main, which doesn't have the updated test permissions yet.  I could have done this as 2 PRs, but it doesn't seem necessary for it to make the real call in tests anyway).

Also fixed a typo in the real file's name, which does call the real URL in tests still, and xfails if it gets a 404 on a test branch, so we can deal with renaming the file in future.


Jongmassey (approved):
Looks good, especially like the improvements to the unit tests!

PR #101, opened 2022-02-08
feat: prompt to update when docker images out of date

bloodearnest:
This is similar to how we prompt for the cli tool itself.

The docker registry API is horrible, and just looking up anything
requires an auth dance, even for public resources.

### DIFF
--- a/opensafely/pull.py
+++ b/opensafely/pull.py
@@ -1,7 +1,10 @@
 import subprocess
 import sys
+from http.cookiejar import split_header_words
 from pathlib import Path
+from urllib.parse import urlparse
 
+from opensafely._vendor import requests
 from opensafely._vendor.jobrunner import config
 from opensafely._vendor.jobrunner.cli.local_run import docker_preflight_check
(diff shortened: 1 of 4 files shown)


opensafely/pull.py
```diff
@@ -119,3 +134,59 @@ def remove_deprecated_images(local_images):
...
                 subprocess.run(["docker", "image", "rm", tag], capture_output=True)
+
+
+session = requests.Session()
+token = None
+
+
+def get_remote_sha(full_name, tag):
+    """Get the current sha for a tag from a docker registry."""
+    global token
```
rebkwok, line 145: Do we need to declare `token` global when it's already been defined outside of the function (line 140)?  It's already a global variable, isn't it?

rebkwok (approved):
just a qn about `global`

opensafely/pull.py
bloodearnest: We do if I want to write to it, which we do if it's not set.

Otherwise, doing `token = ...` will only set a local `token` variable, not update the global one.

PR #106, opened 2022-03-08
Apply Tooling

ghickman:
This applies our standard stack of tooling to the project, applying the usual fixes from those tools.

TODO: pick apart setup.py, pyproject.toml, vendor.in & requirements.*.in files for where dependencies are built from.  This should all be handled in the justfile.


StevenMaude:
This would be good to get done :+1:

ghickman:
Done in #167

PR #110, opened 2022-03-25
feat: add unzip subcommand

bloodearnest:
This simply wraps gunzip as a helpful cross platform tool for users to
be able to examine gzip compressed outputs locally

### DIFF
--- a/opensafely/unzip.py
+++ b/opensafely/unzip.py
@@ -1,0 +1,45 @@
+import gzip
+import sys
+from pathlib import Path
+
+DESCRIPTION = "Unzip a compressesd .csv.gz or .dta.gz file for local viewing"
+
+
+def add_arguments(parser):
+    parser.add_argument(
+        "path", help="path to unzip. If a directory, unzip all .gz files in it"
+    )
+
+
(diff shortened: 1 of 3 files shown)


opensafely/unzip.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        tmp = dst_path.with_suffix(".tmp")
+        with tmp.open("wb") as dst:
+            with gzip.open(src_path) as src:
+                while block := src.read(8192):
+                    dst.write(block)
+    except Exception:
+        tmp.unlink(missing_ok=True)
+    else:
+        tmp.rename(dst_path)
```
evansd, line 44: I think `replace()` is better behaved on Windows, although I can't quite remember the details:
https://docs.python.org/3.8/library/pathlib.html#pathlib.Path.replace

bloodearnest: TIL: https://stackoverflow.com/questions/60369291/os-link-vs-os-rename-vs-os-replace-for-writing-atomic-write-files-what/65553754#answer-65553754

PR #115, opened 2022-04-25
feat: Add a log-stats command

rebkwok:
We now log a bunch of useful stats from cohort-extractor.
This extracts the summary (job id, run by user etc) and cohortextractor-stats from log files and saves them as a JSONL file in the metadata folder.
Intended for use with research-action - see https://github.com/opensafely-core/research-action/pull/49 - which will send the JSONL file to honeycomb


opensafely/__init__.py
```diff
@@ -42,6 +43,7 @@ def add_subcommand(cmd, module):
     add_subcommand("check", check)
     add_subcommand("jupyter", jupyter)
     add_subcommand("unzip", unzip)
+    add_subcommand("log-stats", log_stats)
```
evansd, line 46: If I was going to be _super_ bikesheddy I might say the main verb should be something like "extract" here rather than "log" (e.g. `extract-stats` or `extract-logs`) because that's a more accurate description of what it's doing. I don't hugely care about this, but given that it's awkward to change it later it's maybe worth ~3 minutes of thought.

opensafely/log_stats.py
```diff
@@ -0,0 +1,156 @@
...
+        "action": current_action,
+        **{k: v for k, v in summary_stats.items()},
+    }
+
+
+def main(project_dir, output_file, project_name=None):
+    log_dir = project_dir / "metadata"
+    project_name = project_name or project_dir.resolve().name
+
+    log_files = list((log_dir).glob("*.log"))
```
evansd, line 85: I'm guessing the parens around `log_dir` here are left over from when it used to read `(project_dir / "metadata")`?

opensafely/log_stats.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Find the number of actions (equivalent to the number of log files)
+    action_count = len(log_files)
+    stats_logs = []
+    summary_stats_logs = []
+
+    for filep in log_files:
+        action_count += 1
```
evansd, line 93: This looks weird given that you set `action_count = len(log_files)` already. Did it used to be set to zero?

opensafely/log_stats.py
```diff
@@ -0,0 +1,156 @@
...
+        current_action = filep.stem
+        # Include the total number of actions run alongside this one in each action summary
+        summary_stats = {"total_actions_in_job_request": action_count}
+        raw_logs = []
+        current_log = None
+        for line in filep.open().readlines():
+            # Logs in the log file can span multiple lines;
+            # check if this line is the beginning of a new log or the action summary
+            if re.match(TIMESTAMP_PREFIX_REGEX, line) or re.match(
```
evansd, line 102: Just as an FYI, not something that needs changing, you could define your regex constants using `re.compile()` e.g.
```py
TIMESTAMP_PREFIX_REGEX = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{9}Z")
```
And then later you can use them like:
```py
if TIMESTAMP_PREFIX_REGEX.match(...):
    ...
```
There's no real performance gain these days given that Python caches compiled regexes anyway, but it's syntactically a bit neater.

evansd (approved):
This looks great and I think is going to prove incredibly useful.

I think the only thing that might need addressing before merge is the `action_count` question. But I'll approve as-is and let you worry about that.

Everything else is ignorable bike-shedding .

opensafely/__init__.py
rebkwok: yes, that's better. I wasn't happy with the command nane anyway

opensafely/log_stats.py
rebkwok: indeed :)

opensafely/log_stats.py
rebkwok: yes, it's a leftover; I'll remove it

opensafely/log_stats.py
rebkwok: :+1: I'm never really sure when to use `compile`

evansd: Yeah, it's a stylistic preference really and I didn't want to be imposing mine. Just thought I'd mention it as an option in case in hadn't occurred to you.

opensafely/log_stats.py
```diff
@@ -0,0 +1,156 @@
...
+                COMPLETED_PREFIX_REGEX, line
+            ):
+                # If there's a current log, we're finished with it now, parse it add to the stats dict
```
iaindillingham: I don't understand this comment, I'm afraid.

opensafely/log_stats.py
```diff
@@ -0,0 +1,156 @@
...
+        summary_stats["elapsed_time"] = str(elapsed)
+    return {
+        "timestamp": _timestamp_for_honeytail(start_time, ts_format),
+        "project": project_name,
+        "action": current_action,
+        **{k: v for k, v in summary_stats.items()},
+    }
+
+
+def main(project_dir, output_file, project_name=None):
```
iaindillingham: This comment is going to be unhelpful, but I'm going to make it and reflect on what I can do about it. Anyhow, I found this function _really_ hard to interpret. Partly, that's because I stepped through this PR commit-by-commit, and this function has changed between commits. Partly, that's because I'd not encountered a `cohortextractor-stats` log entry in the wild, so I was missing important context.

tests/fixtures/metadata/action1.log
```diff
@@ -0,0 +1,28 @@
+2022-04-25T11:37:20.751678216Z 2022-04-25 11:37:20 [info     ] Generating cohort for study_definition in output [cohortextractor.cohortextractor] 
+2022-04-25T11:37:20.832847251Z 2022-04-25 11:37:20 [info     ] cohortextractor-stats          [cohortextractor.log_utils] output_column_count=33 table_count=30 table_joins_count=30
```
iaindillingham, line 2: Note to self: skim the whole PR first 🙂

iaindillingham (approved):
Some observations, just to capture my thought process. No action required!

opensafely/log_stats.py
rebkwok: `current_log` is probably a bad name
Logs can span multiple lines; `current_log` is used to keep track of the one that's currently being parsed.  If line we're processing is the beginning of a new timestamped log, or the beginning of the "Completed" section of the log file, we must be done with the previous log, so we add it to the list of stats logs and reset `current_log` to this line.  (Also the comment is wrong, it used to parse it here too, but now it leaves it till the end and parses them all at once when it knows the job id and can add that in too)

opensafely/log_stats.py
rebkwok: I can add more comments :)

iaindillingham: I found it hard to understand intent; I'm still unsure I understand the relationship between `raw_logs` and `current_log`. Am I correct in thinking that `cohortextractor-stats` log entries don't span multiple lines? (But other log entries can.)

Possibly a comment like:

```
Iterate over each line in the log file, collecting cohortextractor-stats log entries in raw_logs.
Finally, parse these log entries into a dict and write to a file.
```

Which isn't quite what's going on, but I think is more-or-less correct 🤔

opensafely/log_stats.py
iaindillingham: Does `current_log` contain the current line or the last line?

opensafely/log_stats.py
rebkwok: I've tried to add more comments to clarify what this is doing.

Both cohortextractor-stats and other logs can span multiple lines.  We want to keep stats logs but ignore other logs.  `current_log` (which is now renamed to `extracted_log`) is keeping track of the log that we're processing; when we read a new line and it's the start of the next log, we have to store the previous log, and start tracking this one (assuming it's a stats log).  So `current_log` isn't the current line when we read a new line, but could be set to the current line when we process it.

That probably doesn't help at all.

opensafely/extract_stats.py
```diff
@@ -78,6 +78,19 @@ def format_summary_stats(project_name, current_action, summary_stats):
 
 
 def main(project_dir, output_file, project_name=None):
+    """
+    Read all log files from the project's metadata folder and extract stats logs.
+
+    Cohort-extractor actions log stats with the event "cohortextractor-stats",
+    which is used to idenitfy stats logs, and any number of key-value pairs of 
+    stats data in the format key=value.  Logs can span multiple lines, and values in
+    a key=value pair can contain spaces.
```
iaindillingham, line 88: All of this commit is great; but this comment is especially great. Thank you 🙂

opensafely/log_stats.py
iaindillingham: The comments are great, thank you 🙂 I hadn't quite appreciated that we're working "one ahead", because of the nature of the log entries.

opensafely/log_stats.py
rebkwok: I am not convinced about this action_count, which in the log I label `total_actions_in_job_request`.  I was thinking about this in the context of logging things about test runs (via `research-action`); in that case, the action will always run `run_all`, and produce one log file per action, so it `total_actions_in_job_request` makes sense.  But if this is run locally, either after a local run with dummy data, or on the server with real log files, it'll always just process all log files it finds and report the total number of them as the number of actions in the job request, when it might not be - if you ran one action but other log files already existed.  I'm not entirely sure how to deal with that, without tying the stats extraction more closely to job-runner.

rebkwok: I could make job-runner log the job request ID as well as the job ID maybe, then it can instead count actions per job request it finds in the logs, rather than just counting all of them

rebkwok:
Another couple of commits added to count actions by job request ID, which job-runner can log (once https://github.com/opensafely-core/job-runner/pull/386 is done, it'll also need the vendored job runner updating)

evansd (approved):

rebkwok:
Bah, updating job-runner broke everything

PR #131, opened 2022-07-27
Enforce single version of codelist

### LINKED ISSUE #130: Misleading error when multiple versions of the same codelist are in `codelists.txt`
@LisaHopcroft reported:

> The codelist check is failing for me when I push to Github:
>
> ```
> It looks like 'codelists.txt' has been edited but 'update' hasn't been run.
>   added: primis-covid19-vacc-uptake/ckd15/v1
>   added: primis-covid19-vacc-uptake/cov1decl/v1.1
>   added: primis-covid19-vacc-uptake/ckd_cov/v1.2.1
>   added: primis-covid19-vacc-uptake/ckd35/v1
>   added: primis-covid19-vacc-uptake/nonshield/v1
>   added: primis-covid19-vacc-uptake/longres/v1
>   added: primis-covid19-vacc-uptake/shield/v1
>   added: primis-covid19-vacc-uptake/learndis/v1
>
> To fix these errors run the command below and commit the changes:
>
>  opensafely codelists update
>
> Error: Process completed with exit code 1.
> ```
>
> ...but running opensafely codelists update doesn't help, I still get the same error.

Lisa later worked out:

> It seems that particular failure was due to different versions of the same codelists in codelists.txt. Could the error message be improved to include this fix suggestion (tech-support)?

https://bennettoxford.slack.com/archives/C0270Q313H7/p1658854715719969

StevenMaude:
Fixes #130.

The reasoning about this is perhaps more complicated than the code.

The problem highlighted in #130 is that codelists that are the same, but have a different version, give a misleading error. The underlying cause is:

1. We gather the codelist IDs from the text file.
2. We gather the codelist IDs in the JSON.
3. With codelists that are identical except for version, in the text file, those are treated as distinct (`codelist.id` is the full path).
4. Codelists that are identical except for version are ignored in the JSON, because the key inside `files` is a `.csv` filename which strips the version. Inside that `.csv` filename key, the object has a key `id` which is the full path with the version. The last codelist found with a given `.csv` key, ends up in the JSON. 
5. We compare the IDs from 1 and 2, and if they are different, the CLI fails and complains that codelists need updating.

It seems like the most sensible place to check is the text file, so I've added the check here.

The same behaviour for handling other errors in the same function is followed: that is, fail at the first failure, instead of bundling up all the errors.


opensafely/codelists.py
```diff
@@ -184,6 +185,13 @@ def parse_codelist_file(codelists_dir):
                 f"{line} does not match [project]/[codelist]/[version] "
                 "or user/[username]/[codelist]/[version]"
             )
+        codelist_id = "/".join(tokens[:-1])
+        if (existing_version := codelist_versions.get(codelist_id)) is not None:
```
StevenMaude, line 189: question: Should we also check for the same version, i.e. redundant duplicated lines? We could here.

StevenMaude:
The question I mainly have is, as I'm more unfamiliar, does this make sense?

Would there ever be a case where you intentionally want two versions of the same codelist?

I suspect not, and this functionality wouldn't have worked in the current state anyway: the keys in the `codelists.json` are a filename, which is stripped of version.

Jongmassey:
> Would there ever be a case where you intentionally want two versions of the same codelist?

Very, very rarely. There is functionality within opencodelists to diff two codelists or versions of the same codelist. One may conceivably want to write a study definition to compare the impact of two versions of a codelist on a study's results. 

> and this functionality wouldn't have worked in the current state anyway

I think that's a clear sign that no-one has wanted to, or at least tried to yet.

opensafely/codelists.py
Jongmassey: I think we should be encouraging cleanliness, and having duplicated lines isn't very clean. I don't think this is critical for acceptance of this change.

Jongmassey (approved):
If you feel compelled to add a warning/error for duplicated codelist entries then that would be nice but otherwise I think it's good

opensafely/codelists.py
StevenMaude: I agree. Though looking again: the current behaviour here actually ends up giving a misleading error — if you have a duplicate, you'll get an error that it's a different version, so I'll revisit this.

StevenMaude:
@Jongmassey: I think this is good now, but please take another look, in case I've done anything silly as it's the end of day.

I've also added an extra test (and checked the errors manually locally; it's not ideal that we don't validate the errors in the tests, but I'm following the existing convention).

Jongmassey (approved):
Covers all bases nicely

PR #144, opened 2022-10-18
update v2.59.10

bloodearnest:
- chore: refactor update tooling
- feat: update job-runner to v2.59.10


evansd (approved):
Man, those `otel` dependencies ...

Do the CLI tests exercise the otel stuff? I'm just wondering whether all those packages can be vendored successfully with no manual tweaks required. Given that jobrunner, ruamel and requests all required tweaking I'd be a bit surprised.

bloodearnest:
It doesn't work, but not because of dependencies, but because of vendoring.

The otel libraries rely on using pkg_resources to inspect the environment they are running in. This doesn't work with our vendoring set up, as the names/metadata doesn't line up

evansd (approved):
So many yaks! This makes sense to me though, and making sure the dependency versions match those in job-runner is obviously a big improvement.

Have you done a full smoke test i.e. install the built package into a fresh venv and tried running a real study with it? If so then I'd be happy for this to ship.

Have fun with the GHA debugging!

bloodearnest:
I have run it on a study with the package installed in the venv, but not in isolation from the built artifcat

Issue #149, opened 2022-10-21
Finish `opensafely exec` implementation


evansd:
The [long-discussed](https://github.com/opensafely-core/opensafely-cli/issues/18) `opensafely exec` command was finally implemented in:
 * #146

But there are further tasks before this is ready for general consumption:
 * Undo the quick fix and implement the proper solution for the "TTY problem" #148
 * Document the `exec` command
 * Rebuild all Docker images from the latest base-docker so they include [this fix](https://github.com/opensafely-core/base-docker/pull/17) which allows them to run as non-root.
 * Check that all images _can_ be run as non-root — they should be able to but there might be some unpleasant surprises that need fixing.
 * Consider the extra features mentioned in the original PR (though we may not want to do any right now):
   - Adding an `--entrypoint` argument to allow customising the image entrypoint for debugging purposes.
   - Passing the appropriate product key for the `stata-mp` image.
   - Setting the `OPENSAFELY_BACKEND` environment variable to `expectations` as the local job-runner does.

StevenMaude:
## Failed to write output

I tested this out. It looks useful but there's one quirk for me at the moment.

When I tried it as is, with the current `opensafely` (v1.32.1) and Data Builder (v0.57.0), it worked without the `--output` option, but not with.

When I try to write a CSV with the `--output` option, with the latest Data Builder image (v0.57.0), I get a permissions issue:

```sh
$ opensafely exec databuilder:v0.57.0 generate-dataset "./1a_minimal_dataset_definition.py" --dummy-tables "example-data/minimal" --output "outputs/1a_minimal_dataset_definition.csv"
```
```python
…
  File "/app/databuilder/file_formats/csv.py", line 16, in write_dataset_csv
    context = filename.open(mode="w", newline="")
  File "/usr/lib/python3.9/pathlib.py", line 1242, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/usr/lib/python3.9/pathlib.py", line 1110, in _opener
    return self._accessor.open(self, flags, mode)
PermissionError: [Errno 13] Permission denied: 'outputs/1a_minimal_dataset_definition.csv'
```

The same command with `opensafely run` *does* work:

```
databuilder:v0.57.0 generate-dataset "./1a_minimal_dataset_definition.py" --dummy-tables "example-data/minimal" --output "outputs/1a_minimal_dataset_definition.csv"
```

[...]

PR #156, opened 2022-10-27
feat: Add toggle for `--user uid:gid` in `opensafely exec`

StevenMaude:
As the `exec` option stands, it doesn't work with rootless Docker.

See https://github.com/opensafely-core/opensafely-cli/issues/149#issuecomment-1293513246

### DIFF
--- a/opensafely/execute.py
+++ b/opensafely/execute.py
@@ -11,6 +11,13 @@ DESCRIPTION = "Run an OpenSAFELY action outside of the `project.yaml` pipeline"
 
 def add_arguments(parser):
     parser.add_argument(
+        "--container-user",
+        dest="container_user_is_enabled",
+        action="store_true",
+        default=False,
+        help="Overrides automatic of host user id and group id or Docker; for rootless Docker",
+    )
+    parser.add_argument(
         "image",
         metavar="IMAGE_NAME:VERSION",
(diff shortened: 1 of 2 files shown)


StevenMaude:
I'm not sure if this is the best solution, but it works for me.

evansd:
Hi @StevenMaude, I'd like to understand a bit more about what's going on here before we make any changes.

Can you confirm the ownership and permissions of the file created with your patch in place? Is it owned by root or by you (or something else)?

In terms of immediately unblocking yourself, you can emulate the current behaviour of `opensafely exec` with a tiny shell script:
```sh
#!/bin/sh
image="$1"
shift 1
exec docker run --rm --volume="$PWD:/workspace" --user "$(id -u):$(id -g)" \
  "ghcr.io/opensafely-core/$image" "$@"
```

If you remove the `--user` argument from that it should do what you want.

lucyb:
Is this PR still needed or should it be closed now?

StevenMaude:
I'll close because you can at least configure the UID:GID in `opensafely exec`.

I think this will still need a wrapper or similar on my own install because there may be cases — like when writing a script — where:

* the default UID and GID won't work because of rootless Docker
* but I don't want to actually override the UID:GID in the script, because it might be intended for others or CI to run

But that's fine; I'll look into it.

PR #167, opened 2022-12-29
Add justfile, and apply all the formatting/linting checks

bloodearnest:


justfile
```diff
@@ -0,0 +1,109 @@
+set dotenv-load := true
+
+# just has no idiom for setting a default value for an environment variable
+# so we shell out, as we need VIRTUAL_ENV in the justfile environment
+export VIRTUAL_ENV  := `echo ${VIRTUAL_ENV:-.venv}`
+
+# TODO: make it /scripts on windows?
+export BIN := VIRTUAL_ENV + "/bin"
+export PIP := BIN + "/python -m pip"
```
ghickman, line 9: The repo-template has an updated version of this which works on windows too.

pyproject.toml
```diff
@@ -55,3 +55,12 @@ drop = [
...
 "ruamel.yaml" = "ruamel"
+
+[tool.isort]
+force_grid_wrap = 0
+include_trailing_comma = true
+line_length = 88
+lines_after_imports = 2
+multi_line_output = 3
+skip_glob = [".direnv", "venv", ".venv", "opensafely/_vendor"]
+use_parentheses = true
```
ghickman, line 66: This can be collapse down using the black profile, [example in the repo-template](https://github.com/opensafely-core/repo-template/blob/main/pyproject.toml#L16-L19)

justfile
```diff
@@ -0,0 +1,109 @@
...
+# upgrade dev or prod dependencies (specify package to upgrade single package, all by default)
+upgrade env package="": virtualenv
+    #!/usr/bin/env bash
+    opts="--upgrade"
+    test -z "{{ package }}" || opts="--upgrade-package {{ package }}"
+    FORCE=true {{ just_executable() }} requirements-{{ env }} $opts
+
+
+test *args: devenv
+    $BIN/coverage run --module pytest {{ args }}
```
ghickman, line 96: We'll also want, on the line below this, `$BIN/coverage report || $BIN/coverage html` otherwise coverage isn't doing very much!

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+  python: python3.10
+
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
+    hooks:
+    - id: black
+
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
```
ghickman, line 11: Could you change this to `6.0.0` to match dev requirements?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
+default_language_version:
+  python: python3.10
+
+repos:
+  - repo: https://github.com/psf/black
+    rev: 22.6.0
```
ghickman, line 6: We've got `21.10b0` in the dev requirements, could you bump both to `22.12.0`? Probably best to start from the current version!

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+    hooks:
+    - id: black
+
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+    - id: flake8
+
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.10.1
```
ghickman, line 16: Could you change this to `5.11.4` to match dev requirements?

ghickman (approved):

justfile
bloodearnest: So, that line is wrong. As it is, it only runs html if report fails.

I think what you want is `$BIN/coverage report || true; $BIN/coverage html`  - that will run both regardless of reports exit code

bloodearnest: Hmm, or is it wrong? 
I think you want to exit with the exit code of `coverage report`, but run html anyway?

PR #169, opened 2022-12-29
Finish off `opensafely exec` command

### LINKED ISSUE #149: Finish `opensafely exec` implementation

bloodearnest:
Finish off the `opensafely exec` implementation.

Fixes #149.

Was a bit involved, because windows, mostly the first commit.

### DIFF
--- a/opensafely/execute.py
+++ b/opensafely/execute.py
@@ -11,37 +11,68 @@ from opensafely.windows import ensure_tty
 DESCRIPTION = "Run an OpenSAFELY action outside of the `project.yaml` pipeline"
 
 
+def get_default_user():
+    try:
+        # On ~unix, in order for any files that get created to have the
+        # appropriate owner/group we run the command using the current user's
+        # UID/GID
+        return f"{os.getuid()}:{os.getgid()}"
+    except Exception:
+        # These aren't available on Windows; but then on Windows we don't have to deal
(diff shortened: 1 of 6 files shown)


evansd (approved):
:tada: Thanks for schlepping through all this stuff. I think this has the potential to make some big usability improvements to opensafely.

bloodearnest:
Arg, sorry, I just have closed this soon, but I wasn't happy with it, so did another take here https://github.com/opensafely-core/opensafely-cli/pull/171

PR #171, opened 2023-01-03
better docker execution

bloodearnest:
- fix: allow exit codes to pass through
- fix: handle possible winpty wrapper in RunFixture
- fix: Add new utils package with run_docker function
- chore: add a test for the jupyter command.
- fix: use run_docker with exec command

### DIFF
--- a/opensafely/__init__.py
+++ b/opensafely/__init__.py
@@ -34,7 +34,7 @@ def main():
         parser.print_help()
         parser.exit()
 
-    parser.set_defaults(function=show_help, handles_unknown_args=False)
+    parser.set_defaults(function=show_help)
     parser.add_argument(
         "--version", action="version", version=f"opensafely {__version__}"
     )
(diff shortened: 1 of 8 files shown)


opensafely/__init__.py
```diff
@@ -81,27 +78,25 @@ def add_subcommand(cmd, module):
...
 
     # if `run`ning locally, run `check` in warn mode
     if function == local_run.main and "format-output-for-github" not in kwargs:
         check.main(continue_on_error=True)
 
-    sys.exit(0 if success is not False else 1)
+    # allow functions to return True/False, or an explicit exit code
+    if success is False:
+        exit_code = 1
+    elif success is True:
```
evansd, line 94: I think this needs `or success is None` to fully replicate the existing behaviour where functions which don't return anything are assumed successful.

opensafely/jupyter.py
```diff
@@ -57,8 +58,12 @@ def add_arguments(parser):
         default=None,
         help="Port to run on",
     )
-    # opt into looser argument parsing
-    parser.set_defaults(handles_unknown_args=True)
+    parser.add_argument(
+        "jupyter_args",
+        nargs=argparse.REMAINDER,
```
evansd, line 63: This is _much_ nicer!

evansd (approved):
:ok_hand:

bloodearnest:
Fixes #149

PR #173, opened 2023-01-05
fix: add PYTHONPATH=/workspace to jupyter invocation

### LINKED ISSUE #160: Put project root directory on `PYTHONPATH` when running `opensafely jupyter`
It would be useful to put the project root directory (the value of `--directory`) on the `PYTHONPATH` when running `opensafely jupyter`, so that scripts can be imported as modules.

Why? [opensafely/appointments-short-data-report](https://github.com/opensafely/appointments-short-data-report) defines `OUTPUT_DIR` in _analysis/\_\_init\_\_.py_. If a script is invoked as a module (`python -m a_script`), then `OUTPUT_DIR` can be accessed with a relative import (e.g. `from .. import OUTPUT_DIR`). However, the same isn't true when running `opensafely jupyter`, because it's not possible to put the project root directory on the `PYTHONPATH`.

bloodearnest:
This allows jupyter notebooks to import from the top level directory of
the workspace code.

Fixes #160

### DIFF
--- a/opensafely/jupyter.py
+++ b/opensafely/jupyter.py
@@ -170,6 +170,9 @@ def main(directory, name, port, no_browser, jupyter_args):
         f"--hostname={name}",
         "--env",
         "HOME=/tmp",
+        # allow importing from the top level
+        "--env",
+        "PYTHONPATH=/workspace",
     ]
 
     debug("docker: " + " ".join(docker_args))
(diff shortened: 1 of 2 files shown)


iaindillingham (approved):
This is great. Thank you 🙂

PR #174, opened 2023-01-05
fix 1.32.4 packaging

bloodearnest:
Fix b0rked packaging from last job-runner update.

Also add deeper packaging test, and tell pre-commit to STFU


.pre-commit-config.yaml
```diff
@@ -6,6 +6,7 @@ repos:
     rev: 22.12.0
     hooks:
     - id: black
+      exclude: opensafely/_vendor/
```
ghickman, line 9: We can put this at the top level to save some repetition: https://pre-commit.com/index.html#pre-commit-configyaml---top-level

ghickman (approved):
All looks sensible to me 👍

evansd (approved):

.pre-commit-config.yaml
bloodearnest: thanks, fixed

PR #182, opened 2023-01-13
feat: Support `opensafely pull databuilder`

### LINKED ISSUE #158: Update `pull` to treat Data Builder as other images
Modify `opensafely  pull` to automatically pull Data Builder images.

The image is currently explicitly excluded:

https://github.com/opensafely-core/opensafely-cli/blob/b3b66217652673ce1800b8971a4859e7343f76a6/opensafely/pull.py#L68-L70

A slight complexity is that we want to use `v0` for all Data Builder images rather than `latest`, so we need a bit of logic to vary the tag in this line:

https://github.com/opensafely-core/opensafely-cli/blob/b3b66217652673ce1800b8971a4859e7343f76a6/opensafely/pull.py#L75

Note that this is a somewhat short-term solution. We'd really like to move to specifying versions for all images, allow those versions to vary between studies and also simplify the surface and implementation of `opensafely pull`. However that's considered too much to bite off as part of the Data Builder work. So this issue is the minimum we can do to 1) allow researchers to easily use Data Builder and 2) set off on the right foot by at least requiring the version of Data Builder to be specified explicitly (everyone will have to specify `databuilder:v0` in their `pipeline.yaml`). This supersedes the discussion in the comments below.

evansd:
This implements the short-term fix as described in #158. When we publish `v1` of Data Builder we will need to update the tag here.

Longer term we'd like to move to a world where _all_ images use version tag rather than `latest`.

Closes #158

### DIFF
--- a/opensafely/pull.py
+++ b/opensafely/pull.py
@@ -65,14 +65,12 @@ def main(image="all", force=False, project=None):
     try:
         updated = False
         for image in images:
-            # currently databuilder is not published, so we ignore it when pulling
-            if image == "databuilder":
-                continue
             tag = f"{REGISTRY}/{image}"
             if force or tag in local_images:
                 updated = True
                 print(f"Updating OpenSAFELY {image} image")
-                subprocess.run(["docker", "pull", tag + ":latest"], check=True)
+                version = get_default_version_for_image(image)
+                subprocess.run(["docker", "pull", tag + f":{version}"], check=True)
 
         if updated:
             print("Cleaning up old images")
@@ -134,9 +132,6 @@ def get_local_images():
             continue
 
         name, sha = line.split("=", 1)
-        # currently databuilder is not published, so we ignore it
-        if "databuilder" in name:
-            continue
         images[name].append(sha)
 
     return images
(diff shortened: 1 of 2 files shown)


opensafely/pull.py
```diff
@@ -65,14 +65,12 @@ def main(image="all", force=False, project=None):
...
-            # currently databuilder is not published, so we ignore it when pulling
-            if image == "databuilder":
-                continue
             tag = f"{REGISTRY}/{image}"
             if force or tag in local_images:
                 updated = True
                 print(f"Updating OpenSAFELY {image} image")
-                subprocess.run(["docker", "pull", tag + ":latest"], check=True)
+                version = get_default_version_for_image(image)
+                subprocess.run(["docker", "pull", tag + f":{version}"], check=True)
```
bloodearnest, line 73: When we do `opensafely pull --project project.yaml` we pull the names and versions of images from the project.yaml.

If someone had `databuilder:tag` in their project, I think this would replace tag with `v0`.

We should probably only do this default tag dance in none project.yaml invocations?

bloodearnest: I now realise the previous code had the same issue, and would have forced `latest`

evansd: Yeah, this is very much the short-term fix. It effectively makes `v0` behave like `latest` for databuilder. At some point before too long we'll bump to `v1` and force the small handful of initial projects off `v0` as that won't be a supported version anymore.

But before we release `v2` of databuilder we'll need to solve this problem properly, however exactly we decide things should work.

As long as the behaviour here is no worse than it currently is for `latest` I think I'm happy.

bloodearnest (approved):

PR #200, opened 2023-03-31
clean

bloodearnest:
- feat: add clean command
- chore: use new clean image functionality in pull command


ghickman (approved):
I've also run this locally it works as expected 👍

PR #223, opened 2023-10-17
Restrict access to Clock Stops and Open Pathways

iaindillingham:
See opensafely-core/ehrql#1646 and opensafely-core/ehrql#1651.

### DIFF
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -55,6 +55,11 @@ RESTRICTED_DATASETS = [
         cohort_extractor_function_names=[],
         ehrql_table_names=["open_prompt"],
     ),
+    RestrictedDataset(
+        name="wl_clockstops",
+        cohort_extractor_function_names=[],
+        ehrql_table_names=["wl_clockstops", "wl_clockstops_raw"],
+    ),
 ]
 
 PERMISSIONS_URL = "https://raw.githubusercontent.com/opensafely-core/opensafely-cli/main/repository_permissions.yaml"
--- a/opensafely/check.py
+++ b/opensafely/check.py
@@ -60,6 +60,11 @@ RESTRICTED_DATASETS = [
         cohort_extractor_function_names=[],
         ehrql_table_names=["wl_clockstops", "wl_clockstops_raw"],
     ),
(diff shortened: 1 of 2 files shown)


tests/fixtures/permissions/repository_permissions.yaml
```diff
@@ -12,5 +12,9 @@ opensafely/dummy_isaric:
   allow: ['isaric']
 opensafely/dummy_open_prompt:
   allow: ['open_prompt']
+opensafely/dummy_wl_clockstops:
+  allow: ['wl_clockstops']
+opensafely/dummy_wl_openpathways:
+  allow: ['wl_openpathways']
 opensafely/dummy_all:
-  allow: ['icnarc', 'ons', 'isaric', 'ons_cis', 'ukrr', 'open_prompt']
+  allow: ['icnarc', 'ons', 'isaric', 'ons_cis', 'ukrr', 'open_prompt', 'wl_clockstops', 'wl_openpathways']
```
rebkwok, line 20: Should we take ons/ons_cis out of the tests too?

opensafely/check.py
```diff
@@ -41,7 +41,7 @@ class RestrictedDataset:
         cohort_extractor_function_names=[
             "with_an_ons_cis_record",
         ],
-        ehrql_table_names=["ons_cis_raw", "ons_cis"],
+        ehrql_table_names=[],
```
rebkwok, line 44: Should we remove the entire `RestrictedDataset`?  The function `with_an_ons_cis_record` no longer exists in cohort-extractor; if we're removing the ehql references, we probably should do the same for cohort-extractor

rebkwok (approved):

tests/fixtures/permissions/repository_permissions.yaml
iaindillingham: We should! Thank you for spotting this.

opensafely/check.py
iaindillingham: We should! I hadn't realized -- or more likely, I had forgotten -- that we removed ONS-CIS from cohort-extractor too.

PR #226, opened 2023-10-25
Add a new codelists check-upstream command

### LINKED ISSUE #225: Add a command to check codelists are up to date with OpenCodelists
Part of the plan to [address acute codelist rot in ehrQL](https://github.com/opensafely-core/ehrql/issues/1581).

Add a command to check codelists are up to date with OpenCodelists.  dm+d codelists are now downloaded with any mapped VMPs included in the CSV download.  This means that dm+d codelists can become out of date after they've been downloaded into the study, even though there's no new codelist version.

Depends on https://github.com/opensafely-core/opencodelists/issues/1702

rebkwok:
Fixes #225 
Part of the [plan for dealing with acute codelist rot](https://github.com/orgs/opensafely-core/projects/13/views/1?pane=issue&itemId=42536167)

OpenCodelists now has an endpoint that takes the content of a study repo's codelists.txt and codelists.json, and checks that the data that would be downloaded for the specified codelists is the same as the sha in the codelists.json file.  This is specifically intended to deal with changing VMPs in dm+d codelists, in case new mappings have been introduced since the codelist was downloaded into the study repo.

The new subcommand `check-upstream` just calls the opencodelists endpoint, and can be run on its own.  The existing `opensafely codelists check` command now also runs `check-upstream`, so `check` is a comprehensive check of both local changes that may have occurred, and upstream changes.

If any connection error is encountered during the upstream check, we report it in the output message, but don't fail the check.

### DIFF
--- a/tests/test_codelists.py
+++ b/tests/test_codelists.py
@@ -69,7 +69,7 @@ def codelists_path(tmp_path):
 def test_codelists_check(mock_check, codelists_path):
     mock_check(response={"status": "ok"})
     os.chdir(codelists_path)
-    codelists.check()
+    assert codelists.check()
 
 
 def test_codelists_check_passes_if_opencodelists_is_down(requests_mock, codelists_path):
@@ -78,7 +78,7 @@ def test_codelists_check_passes_if_opencodelists_is_down(requests_mock, codelist
         exc=requests.exceptions.ConnectionError,
     )
     os.chdir(codelists_path)
-    codelists.check()
+    assert codelists.check()
 
 
 def test_codelists_check_fail_if_list_updated(codelists_path):
@@ -191,7 +191,7 @@ def test_codelists_check_upstream_fails_if_no_manifest_file(tmp_path, mock_check
 def test_codelists_check_upstream(codelists_path, mock_check):
     mocked = mock_check(response={"status": "ok"})
     os.chdir(codelists_path)
-    codelists.check_upstream()
+    assert codelists.check_upstream()
 
     # assert content sent to opencodelists
     assert mocked.called_once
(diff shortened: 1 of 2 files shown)


tests/test_codelists.py
```diff
@@ -53,7 +66,17 @@ def codelists_path(tmp_path):
     yield tmp_path
 
 
-def test_codelists_check(codelists_path):
+def test_codelists_check(mock_check, codelists_path):
+    mock_check(response={"status": "ok"})
+    os.chdir(codelists_path)
+    codelists.check()
```
iaindillingham, line 72: Why isn't there an `assert` here?

tests/test_codelists.py
```diff
@@ -53,7 +66,17 @@ def codelists_path(tmp_path):
...
+
+
+def test_codelists_check_passes_if_opencodelists_is_down(requests_mock, codelists_path):
+    requests_mock.post(
+        "https://www.opencodelists.org/api/v1/check/",
+        exc=requests.exceptions.ConnectionError,
+    )
     os.chdir(codelists_path)
     codelists.check()
```
iaindillingham, line 81: I appreciate that this isn't touched by your PR, but I think there should be an `assert` here, too.

iaindillingham (approved):
Couple of questions about `assert`s, but that's all.

tests/test_codelists.py
rebkwok: Because I copied it from another test that didn't assert?  If there's a problem, `codelists.check()` will raise an error, so it doesn't need an _assert_ in order to fail the test. I  guess it could have one, just to be more explicit about it

PR #229, opened 2023-11-13
fix: Return False if upstream codelist check fails in CI

rebkwok:
Makes `opensafely codelists check` print the error and return False (but not exit with an error) when running in CI, to avoid failing the entire GH "test that the project is runnable" action for study repos.

### DIFF
--- a/opensafely/codelists.py
+++ b/opensafely/codelists.py
@@ -135,18 +135,29 @@ def check_upstream(codelists_dir=None):
         # if any codelists in codelists.csv don't match the expected pattern. These should all
         # be fixable by `opensafely codelists update`.
         if "error" in response["data"]:
-            exit_with_prompt(
+            error_message = (
                 f"Error checking upstream codelists: {response['data']['error']}\n"
             )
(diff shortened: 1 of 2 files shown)


opensafely/codelists.py
```diff
@@ -210,14 +221,16 @@ def check():
...
-        return check_upstream(codelists_dir)
+        upstream_check = check_upstream(codelists_dir)
     except requests.exceptions.ConnectionError:
         print(
             f"Local codelists OK; could not contact {OPENCODELISTS_BASE_URL} for upstream check,"
             "try again later"
         )
         return True
 
+    return upstream_check
```
evansd, line 232: I think returning `False` here will cause the entire command to exit with an error. See:
https://github.com/opensafely-core/opensafely-cli/blob/b87bef3900a858959bbd19099c770c4730e38668/opensafely/__init__.py#L136-L142

Which I think is not the behaviour you wanted.

rebkwok: No, it's not, thanks.

evansd (approved):

PR #230, opened 2023-11-14
feat: bump job-runner to 2.72, which checks level 4 files local run

bloodearnest:


evansd (approved):
I don't actually know what `RECORD` files are or why the hashes in some of those would have changed. I can't see that it matters though, the actual code itself for these packages hasn't changed.

bloodearnest:
> I don't actually know what `RECORD` files are or why the hashes in some of those would have changed. I can't see that it matters though, the actual code itself for these packages hasn't changed.

I think its the hash of the metadata, which I think has a timestamp in it somewhere, hence the churn

PR #239, opened 2023-12-11
fix: support multiple versions in pull command

bloodearnest:
- Add opensafely label in image filters.
- Support v2 python in opensafely pull

### DIFF
--- a/opensafely/pull.py
+++ b/opensafely/pull.py
@@ -1,6 +1,6 @@
+import argparse
 import subprocess
 import sys
-from collections import defaultdict
 from http.cookiejar import split_header_words
 from pathlib import Path
 from urllib.parse import urlparse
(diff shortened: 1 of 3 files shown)


opensafely/pull.py
```diff
@@ -104,19 +125,22 @@ def get_local_images():
...
     for line in ps.stdout.splitlines():
         if not line.strip():
             continue
 
-        name, sha = line.split("=", 1)
-        images[name].append(sha)
+        line = line.replace("ghcr.io/opensafely-core/", "")
+
+        image, sha = line.split("=", 1)
+        assert image not in images
```
ghickman, line 142: Is this likely to happen?  If so can we throw something more meaningful for the user?  I'm thinking a note to contact tech support is likely more than enough!

opensafely/pull.py
```diff
@@ -133,6 +157,8 @@ def remove_deprecated_images(local_images):
 def get_default_version_for_image(name):
     if name in ["databuilder", "ehrql"]:
         return "v0"
+    elif name == "python":
+        return "v2"
     else:
         return "latest"
```
ghickman, line 163: You could absolutely use a case statement here!  It would improve just about nothing but you _could_ 😜

ghickman (approved):

opensafely/pull.py
bloodearnest: Sadly not, as opensafely-cli runs on user's installed python, so we have to support old python (current lower bound is 3.9, IIRC)

opensafely/pull.py
bloodearnest: It's not at all likely to happen. I threw it in to check it wasn't, I'll remove it, thanks

opensafely/pull.py
ghickman: booooo (but totally understandable)

PR #257, opened 2024-05-07
add another appointments data exception

alexwalkerepi:
I've previously spoken to the researcher about their use of the data and think it's fine - specifically, they're using the data as a variable in a model (as a proxy for healthcare usage), and not as an outcome variable.

### DIFF
--- a/repository_permissions.yaml
+++ b/repository_permissions.yaml
@@ -42,3 +42,5 @@ opensafely/deaths-at-home-covid19:
   allow: ['appointments']
 opensafely/end-of-life-carequality:
   allow: ['appointments']
+opensafely/post-covid-kidney-outcomes:
+  allow: ['appointments']


alexwalkerepi:
> Could you possibly document _why_ you think it's fine, @alexwalkerepi? I imagine we'll revisit this file, and for each exception ask this question. If there's nothing in the commit, PR, or Slack, then we won't know.

Thanks, have edited the description (and also added a `feat` tag that I once again forgot).

iaindillingham:
Turns out you don't need the "feat" prefix, as the list of exceptions is fetched from GitHub.

alexwalkerepi:
> Turns out you don't need the "feat" prefix, as the list of exceptions is fetched from GitHub.

Oh no, now I have to sometimes remember to do it and sometimes not depending what I'm making a PR about :dizzy_face:

PR #264, opened 2024-08-21
fix: Ignore dangling images in opensafely pull

bloodearnest:
The primary purpose of this PR is to ignore dangling images, and thus un-break
`opensafely pull` for some users.

Also included is a related UX improvement to the `opensafely info` command, as
this would have helped diagnose the problems.

This also involved some clean up refactoring, moving the responsibility for
printing warnings to where it belongs.


evansd (approved):

evansd:
Needs a pedantic commit prefix!

PR #266, opened 2024-08-21
fix: older docker doesn't support dangling=false filter

bloodearnest:
- **fix: attempt to fix release action**
- **fix: manually filter for dangling images for older dockers**
- **Make info command throw exceptions rather than handle them**

### DIFF
--- a/.github/workflows/build_and_publish.yaml
+++ b/.github/workflows/build_and_publish.yaml
@@ -57,7 +57,7 @@ jobs:
         python setup.py sdist bdist_wheel
 
     - name: Publish package
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@ec4db0b4ddc65acdf4bff5fa45ac92d78b56bdf0  # v1.90
       if: needs.tag-new-version.outputs.tag
       with:
         user: __token__
--- a/opensafely/info.py
+++ b/opensafely/info.py
@@ -1,6 +1,5 @@
 import json
 import subprocess
-import sys
 
 import opensafely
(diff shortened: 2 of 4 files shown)


.github/workflows/build_and_publish.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
         python setup.py sdist bdist_wheel
 
     - name: Publish package
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@ec4db0b4ddc65acdf4bff5fa45ac92d78b56bdf0
```
evansd, line 60:
```suggestion
      uses: pypa/gh-action-pypi-publish@ec4db0b4ddc65acdf4bff5fa45ac92d78b56bdf0 # v1.90
```
A pattern we've used elsewhere is to add the version tag as a comment. Dependabot is clever enough to understand this and keeps the comment up-to-date when it bumps the version.

opensafely/info.py
```diff
@@ -38,15 +37,15 @@ def main():
                 cpu=info["NCPU"],
             )
         )
-    except Exception:
-        sys.exit("Error retreiving docker information")
+    except Exception as exc:
+        raise Exception(f"Error retreiving docker information: {exc}")
```
evansd, line 41: I guess there's nothing strictly wrong with it but raising `Exception` looks weird to my eyes because I only every see it used as a base class. I tend to use `RuntimeError` as the most generic "a thing went wrong" exception if I'm raising something.

evansd (approved):

Issue #277, opened 2024-10-09
Proposal: make `uv` the documented and supported way to install `opensafely-cli`


evansd:
## Background

We currently distribute the OpenSAFELY user tooling via PyPI. Our users often aren't Python users, and are running a variety of different operating systems with different levels of administration privileges, and have therefore installed Python in a variety of different ways. To cope with this we do two things:
 * we test the tooling on a range of Python version (currently 3.8–3.10);
 * we ensure that the tooling vendors all its dependencies so it can be installed in any environment without risk of interaction with other packages in that environment.

This has some annoying consequences:
 * our code can't use features from newer Python versions;
 * we're [limited](https://github.com/opensafely-core/pipeline/issues/340) to only using pure Python dependencies because only those admit cross-platform vendoring;
 * updating dependencies is more faffy and difficult because of the vendoring (sometimes requiring [manual surgery](https://github.com/opensafely-core/opensafely-cli/blob/d9634cb79b877b32e1a8f6fe9b0b00f2d46dedbe/pyproject.toml#L22-L42)).

_Even worse_, because the CLI tools depends on the `pipeline` library (for parsing `project.yaml` files) and `job-runner` (for running jobs locally) these limitations are transitively imposed on those libraries as well.

And, despite all this pain on our side, the user installation story still isn't great here because users still have to wrestle with installing Python somehow.

We've considered a variety of more drastic solutions here (e.g. build and distribute our own application which bundles Python itself; rewrite everything in Go) but not made any progress so far.

## How might `uv` help?

`uv` is a tool for installing and managing Python and Python packages:
https://docs.astral.sh/uv/

It has a reasonably nice installation story across all platforms relevant to us:
https://docs.astral.sh/uv/getting-started/installation/

Once `uv` is installed it can take care of installing the appropriate version of Python and then installing our tooling and its dependencies in an isolated environment. This means we can require a specific, up-to-date version of Python and no longer need to vendor our dependencies.

Compared to the status quo there are literally no downsides. But there will inevitably be a switching cost in getting existing users to move across, and there are still some unanswered questions.

## Unanswered questions

1. **Is `uv` actually easy to install in the contexts are users are operating in?**
   I think the first thing we should do is get some friendly users (spread across a variety of platforms) to try to install it, based purely on the docs `uv` provide.

2. **Do we want to use `uvx` or `uv tool install`?**
[...]

PR #278, opened 2024-10-15
Make it possible to install with `uv`

evansd:
As a first step in exploring the `uv` story we make the minimal changes necessary to ensure that everything works correctly under `uv` with a recent version of Python. This involves:
 * making explicit our implicit dependency on `setuptools`;
 * prompting the user to upgrade by running `uv tool upgrade opensafely` rather than `opensafely upgrade` if they've used `uv` rather than `pip` to install it.

These changes ought to be invisible to existing users, but they mean it's now possible to get some users to try out `uv` on an experimental basis.

There's very little change in the production code here: as you might expect, most of the shenanigans are with the tests.

Relates to #277

### DIFF
--- a/tests/test_packaging.py
+++ b/tests/test_packaging.py
@@ -6,14 +6,31 @@ from pathlib import Path, PurePath
 
 import pytest
 
+import opensafely
+
 
 BIN_DIR = "bin" if os.name != "nt" else "Scripts"
 
 project_fixture_path = Path(__file__).parent / "fixtures" / "projects"
 
 
+@pytest.fixture
+def older_version_file():
+    # This is really not very nice, but short of reworking the way versioning is handled
+    # (which I don't want to do at the moment) I can't think of another way. In order to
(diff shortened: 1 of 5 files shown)


tests/test_packaging.py
```diff
@@ -6,14 +6,31 @@
...
+import opensafely
+
 
 BIN_DIR = "bin" if os.name != "nt" else "Scripts"
 
 project_fixture_path = Path(__file__).parent / "fixtures" / "projects"
 
 
+@pytest.fixture
+def version_file():
```
bloodearnest, line 18: `older_version_file` for a name perhaps?

bloodearnest (approved):
LGTM, glad to see some movement on this

tests/test_packaging.py
evansd: Oh yeah, that's better. I knew this name wasn't quite right!

PR #282, opened 2024-10-21
fix: test and fix error handling in open_browser

bloodearnest:
This fixes an ugly message for a user when something didn't work right.

There were two errors:
 - a missing default definition of `response`
 - the special formatting of exceptions was broken.

These were not covered by tests, as it was awkward to do. But its caused
breakage, so this PR grasps the nettle and adds coverage for:

 - rstudio command opening the browser
 - jupyter command getting metadata and opening the browser
 - `open_browser` error cases

This required some refactoring of the SubprocessRunFixture, and some other
clean ups.

As a result, test coverage went up by 6%.

- **Refactor SubprocessRunFixture to handle concurrent declarations**
- **Explicitly mock open_browser call in test_rstudio.py**
- **Fully mock/test jupyter browser calls**
- **use requests rather than urllib, as its easier to use**
- **fix: fix issue in exception rendering refactor**
- **Refactor threaded debugging to work for testing**
- **Test and fix the open_browser error handling**
- **bump coverage minimum up 6 to 88**

### DIFF
--- a/tests/test_jupyter.py
+++ b/tests/test_jupyter.py
@@ -43,7 +43,7 @@ def test_jupyter(run, no_user, monkeypatch):
             "foo",
         ]
     )
-    # fetach the metadata
+    # fetch the metadata
     run.expect(
         [
             "docker",
--- a/tests/test_jupyter.py
+++ b/tests/test_jupyter.py
@@ -1,10 +1,19 @@
 import pathlib
+from unittest import mock
 
-from opensafely import jupyter
+from opensafely import jupyter, utils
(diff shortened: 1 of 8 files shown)


tests/test_jupyter.py
```diff
@@ -34,5 +43,18 @@ def test_jupyter(run, no_user):
             "foo",
         ]
     )
+    # fetach the metadata
```
remlapmot, line 46: Typo in comment: amend 'fetach' to 'fetch'

evansd (approved):

PR #294, opened 2024-11-28
`codelists add` command

### LINKED ISSUE #295: UX for adding codelists to a project is poor
To add a codelist from OpenCodelists to a study a user currently has to:
1. Identify the codelist required on OpenCodelists
1. Copy the part of the URL after `https://www.opencodelists.org/codelist/` being sure not to include the trailing `/`
1. Paste into `codelists/codelists.txt`
1. Run `opensafely codelists update`
1. Note the path of the codelist csv that has been downloaded by this command
1. Inspect the csv/the codelist on OpenCodelist to note the key field name(s)
1. Use the `codelist_from_csv()` function in their ehrQL with the field name(s) and csv file path

That's rather a lot of steps, and opportunities for things to go wrong.

Improvement of 5-7 could require changes to ehrQL or perhaps generation of code snippets, but I feel 2-4 are ripe for some simplification by changing the behaviour of the CLI.

Jongmassey:
Allow users to add OpenCodelists codelists to the codelists.txt file by URL.

Fixes #295


opensafely/codelists.py
```diff
@@ -63,7 +68,32 @@ def main():
...
+
+    # turn a download url into a base codelist version url
+    if line.endswith(".csv"):
+        line = line[:line.rfind("/")]
+
+    with codelists_file.open("r+") as f:
+        if not f.readlines()[-1].endswith("\n"):
+            line = "\n" + line
+        f.write(line + "\n")
+    update(codelists_dir, delete_unknown=False)
```
lucyb, line 93: Calling `update(..)` here seems like it's going to cause some unwanted side effects. I can see you've guarded against one unintended side effect by changing the behaviour of `update(..)` when called by `add(..)`, but I'm not sure that users would expect or want their other codelists to change just because they've added an extra one.

Could you move the relevant code out of `update(..)` into a separate function, which is called by both `update(..)` and `add(..)` instead? This would remove the need to change the behaviour of `update(..)`.

tests/test_codelists.py
```diff
@@ -260,3 +260,87 @@ def test_codelists_check_with_upstream_changes_in_CI(
     os.chdir(codelists_path)
     # check doesn't fail in CI if there are upstream errors only
     assert codelists.check()
+
+
+def test_codelists_add(codelists_path, requests_mock):
+    codelists_path /= "codelists"
+    codelists_file = codelists_path / "codelists.txt"
+    prior_codelists = codelists_file.read_text()
+    for codelist in prior_codelists.split("\n"):
```
lucyb, line 269: Using `splitlines()` rather than `split("\n")` will remove the need for the `if codelist` check on the following line.

opensafely/codelists.py
```diff
@@ -123,8 +123,13 @@ def update(codelists_dir=None):
     preserve_download_dates(manifest, manifest_file)
     manifest_file.write_text(json.dumps(manifest, indent=2))
     for file in old_files - new_files:
-        print(f"Deleting {file.name}")
-        file.unlink()
+        if delete_unknown:
```
lucyb: The "Moderate update's' destructive behaviour" commit is no longer needed for this PR to work, I think? Could you move it to a separate one, so we can think about it separately, please? It's a behaviour change to an existing piece of functionality, so it's probably easiest all round if we avoid letting it interfere with this change.

tests/test_codelists.py
```diff
@@ -260,3 +260,87 @@ def test_codelists_check_with_upstream_changes_in_CI(
...
+        if codelist:
+            requests_mock.get(
```
lucyb, line 271: Actually, I can comment out the call to requests_mock and the whole for loop and it still works (for all three tests). It's only the mock of the specific version that is needed. I assume this is a side-benefit of no longer relying on `update(..)`.

opensafely/codelists.py
```diff
@@ -63,39 +68,91 @@ def main():
...
 
+def add(codelist_url, codelists_dir=None):
+    if not codelists_dir:
+        codelists_dir = Path.cwd() / CODELISTS_DIR
+
+    codelists_file = get_codelist_file(codelists_dir)
+
+    if OPENCODELISTS_BASE_URL not in codelist_url:
+        exit_with_error(f"Unable to parse URL, {OPENCODELISTS_BASE_URL} not found")
+    line = codelist_url.replace(f"{OPENCODELISTS_BASE_URL}/codelist/", "")
```
lucyb, line 79: `line` is not a very descriptive name. Is it a codelist descriptor, identifier, name or path, or something else?

opensafely/codelists.py
```diff
@@ -63,39 +68,91 @@ def main():
     pass
 
 
+def add(codelist_url, codelists_dir=None):
+    if not codelists_dir:
+        codelists_dir = Path.cwd() / CODELISTS_DIR
+
+    codelists_file = get_codelist_file(codelists_dir)
+
+    if OPENCODELISTS_BASE_URL not in codelist_url:
```
lucyb, line 77: Can we fail earlier with this check? I guess we should have a test for it as well?

lucyb (commented):
Sorry, I haven't quite finished the review, but I thought some feedback was better than none.

It's looking much better, I can even see what the update function is doing! Thanks for making that bit of refactoring. I think it's nearly done now.

opensafely/codelists.py
Jongmassey: Yes, I plan to squash it out of existence, sorry I should have made that more clear. I left it intact for a point of comparison whilst reviewing

lucyb: You could move it into a separate branch and raise another PR for it. It sounds like it might be useful behaviour to change, but I think I'd prefer to get some feedback from others first. Alternatively, we could just repeat the work another time..

tests/test_codelists.py
Jongmassey: I copied this from one of the other tests, but I shall remember this useful distinction in future!

opensafely/codelists.py
Jongmassey: You're right. I'd copied this convention from the previous code but it's an opportunity to make it better.

opensafely/codelists.py
Jongmassey: It certainly could form part of a rethink of what the `update` sub-command means, but it doesn't really make any sense in isolation. I'd rather not move it to its own draft PR as not to incur the wrath of the "long running PRs" metric.

opensafely/codelists.py
Jongmassey: Yes, and yes!

tests/test_codelists.py
Jongmassey: Ah yes, I hadn't spotted that in the refactoring. You're entirely right, this call is no longer needed without the call to `update()` so I've removed it.

opensafely/codelists.py
```diff
@@ -63,39 +68,92 @@ def main():
     pass
 
 
+def add(codelist_url, codelists_dir=None):
+    if OPENCODELISTS_BASE_URL not in codelist_url:
+        exit_with_error(f"Unable to parse URL, {OPENCODELISTS_BASE_URL} not found")
+
+    if not codelists_dir:
```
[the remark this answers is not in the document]
Jongmassey: ah yes good spot, I've changed to make use of this

tests/test_codelists.py
```diff
@@ -260,3 +260,88 @@ def test_codelists_check_with_upstream_changes_in_CI(
...
+        == prior_codelists + "project123/codelist456/version1\n"
+    )
+    assert (codelists_path / "project123-codelist456.csv").read_text() == "foo"
+
+
+def test_codelists_add_with_anchor_url(codelists_path, requests_mock):
+    codelists_path /= "codelists"
+    codelists_file = codelists_path / "codelists.txt"
+    prior_codelists = codelists_file.read_text()
+    for codelist in prior_codelists.split("\n"):
```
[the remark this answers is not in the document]
Jongmassey: Correct, the second commit (I know in retrospect it seems a bit backwards now!) makes the mock responses for the prior codelists not needed. However the mock response for the codelist being added is still needed to make sure that the correct URL is requested and that the csv contents are correctly written.

tests/test_codelists.py
```diff
@@ -260,3 +260,88 @@ def test_codelists_check_with_upstream_changes_in_CI(
...
+        == prior_codelists + "project123/codelist456/version1/\n"
+    )
+    assert (codelists_path / "project123-codelist456.csv").read_text() == "foo"
+
+
+def test_codelists_add_with_download_url(codelists_path, requests_mock):
+    codelists_path /= "codelists"
+    codelists_file = codelists_path / "codelists.txt"
+    prior_codelists = codelists_file.read_text()
+    for codelist in prior_codelists.split("\n"):
```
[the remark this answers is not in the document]
Jongmassey: Correct, the second commit (I know in retrospect it seems a bit backwards now!) makes the mock responses for the prior codelists not needed. However the mock response for the codelist being added is still needed to make sure that the correct URL is requested and that the csv contents are correctly written. I've updated to remove the no-longer-needed bits

Jongmassey:
Good job you asked about running the CLI tool, I did so and found that the `codelists` sub parser was missing the `codelist_url` arg ([now fixed](https://github.com/opensafely-core/opensafely-cli/pull/294/commits/4ec9780a36a49570d33ae50347e6236742293fd1#diff-7557cda401c50f2b44a965eaec6aff7a1ad0379a7dddff6ccda2b0595f8e4a1cR35-R36)).

I've also re-ordered the commits so hopefully they make more sense

PR #301, opened 2025-01-30
Add new launch command, deprecated rstudio/jupyter commands

bloodearnest:
Add a new launch command and deprecate rstudio/jupyter commands.

The launch command is designed to run a server, and open a browser pointing to
it.  The rstudio and jupyter commands did this previously, now becoming
`opensafely launch rstudio` and `opensafely launch jupyter` repectively.

Putting both command because a generic launc tool has several benefits:
 - they are both similar, concentrating and sharing code
 - it allows us to easily use `opensafely launch tool:version`, which has two benefits
   - similar to `opensafely exec image:version`
   - saves a lot of mess with teaching argparse to handle the :
   - can add more tools easily in future.

There's a few small changes along the way, harmonizing the arguments between
them mainly, and fixing launching jupyter on python:v2 image.


opensafely/launch.py
```diff
@@ -0,0 +1,207 @@
...
+        default=False,
+        action="store_true",
+        help="Do not attempt to open a browser",
+    )
+
+
+def add_arguments(parser):
+    parser.add_argument(
+        "tool",
+        help="Tool to launch e.g. rstudio",
```
rebkwok, line 45: This could have a `choices` arg? Then it shouldn't need to raise the `ArgumentTypeError` because it would only accept rstudio/jupyter options

opensafely/jupyter.py
```diff
@@ -1,130 +1,13 @@
...
-
-    ps = utils.run_docker(
-        docker_args, "python", jupyter_cmd, interactive=True, directory=directory
+    print(
+        "opensafely jupyter command is deprecated - instead use: opensafely launch jupyter"
     )
-
-    # we want to exit with the same code that jupyter did
-    return ps.returncode
+    return launch.main("jupyter", directory, name, port, no_browser)
```
rebkwok: I wonder if it's better to just return after the deprecation message so users have to use the new command and get used to it

opensafely/launch.py
rebkwok: Actually, now I see that we still need that check in `main` because the deprecated commands are calling it directly. Adding choices here would still give you a nicer help text that tells you what the options are though

rebkwok (approved):
Couple of questions, nothing blocking

opensafely/launch.py
bloodearnest: Good idea

opensafely/jupyter.py
bloodearnest: Also not a bad idea, will try  it

opensafely/launch.py
bloodearnest: Ah, I remember why I didn't to this.

I don't want to have to pre-enumerate all the versions, i.e. 
```choices=["jupyter", "jupyter:v1", "jupyter:v2", "rstudio", "rstudio:v1", "rstudio:v2", ...]
```

I'd like to to work with new versions of image w/o updates.

My custom validation effectively checks that the name before the `:` is valid. You could make argparse do this, but I think it ends up more complex.

PR #302, opened 2025-02-04
feat: add --background and --force options to launch command

bloodearnest:
To use `opensafely launch rstudio` in codespaces, we need to run in the background.  This is more complex than just adding `&` to the invocation, as docker client is actually forwarding stdin/out/err from the docker server - its pretending to be a process we control, but its not really (another reason we should look at podman). We need to use `docker run -d`. So this change adds a --background flag.  This is actually really useful for testing, and allows us to add fully functional tests of `opensafely launch` that were very tricky to write when running in the foreground.

In addition, it would be helpful to make launch command idempotent, so if you re-run it, you reconnect to the old running instance, which will preserve state between codespace use and provide better UX.  We add a --force to allow users recreate the session, which is maybe useful to ensure they are running the latest version of the image if needed.

There's also a few cleanups that made this work better, mainly bringing jupyter and rstudio implementations closer together, and removing some unneeded complexity.

### DIFF
--- a/opensafely/launch.py
+++ b/opensafely/launch.py
@@ -126,8 +126,7 @@ def launch_jupyter(version, directory, name, port, no_browser, background):
         "--Application.log_level=ERROR",  # only log errors
     ]
 
-    print(f"Running following jupyter cmd in OpenSAFELY docker container {name}...")
-    print(" ".join(jupyter_cmd))
+    utils.debug(" ".join(jupyter_cmd))
 
     docker_args = [
         # we use our port on both sides of the docker port mapping so that
@@ -206,7 +205,6 @@ def launch_rstudio(version, directory, name, port, no_browser, background):
         directory=directory,
     )
 
-    utils.debug("docker: " + " ".join(docker_args))
     print(f"Opening an RStudio Server session at {url}.")
     return run_tool(docker_args, kwargs, background, no_browser, url)
 
--- a/opensafely/launch.py
+++ b/opensafely/launch.py
@@ -36,6 +36,13 @@ def add_base_args(parser):
         action="store_true",
         help="Do not attempt to open a browser",
     )
+    parser.add_argument(
+        "--background",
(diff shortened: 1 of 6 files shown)


tests/test_launch.py
```diff
@@ -136,3 +136,32 @@ def test_rstudio(run, tmp_path, monkeypatch, gitconfig_exists, version):
 
     assert run_main(launch, f"{tool} --port 8787 --name test_rstudio") == 0
     mock_open_browser.assert_called_with("http://localhost:8787")
+
+
+@pytest.mark.skipif(sys.platform != "linux", reason="Only runs on Linux")
+@pytest.mark.parametrize("tool", ["jupyter", "rstudio"])
+@pytest.mark.functional
+def test_launch_browser(monkeypatch, docker, tool):
+    """This test's the --background flag, as well as providing base functional tests."""
```
rebkwok, line 140: nit s/test's/tests

opensafely/launch.py
```diff
@@ -119,10 +119,11 @@ def launch_jupyter(version, directory, name, port, no_browser, background):
         "lab",
         "--ip=0.0.0.0",
         f"--port={port}",
-        "--allow-root",
-        "--no-browser",
+        "--no-browser",  # we open the brower
```
rebkwok, line 121: s/brower/browser

opensafely/launch.py
```diff
@@ -119,10 +119,11 @@ def launch_jupyter(version, directory, name, port, no_browser, background):
...
         # display the url from the hosts perspective
-        f"--LabApp.custom_display_url=http://localhost:{port}/",
+        f"--ServerApp.custom_display_url=http://localhost:{port}/",
+        "-y",  # do not ask for confirmation on quiting
```
rebkwok, line 124: s/quiting/quitting

rebkwok (approved):

PR #313, opened 2025-03-05
opensafely launch tweaks for better codespaces UX

bloodearnest:
This PR makes a couple of changes to make the experience of running opensafely
launch rstudio:v2 work better in codespaces.

Specifically, re-using the same port and showing the initial pull to user.


opensafely/utils.py
```diff
@@ -182,16 +182,27 @@ def dockerctl(cmd, *args, check=True, capture_output=True, text=True, **kwargs):
     )
 
 
-def get_free_port():
+def get_free_port(default=None):
     """Get a port that is free on the users host machine"""
+    # try the default first
+    if default:
```
lucyb, line 188: Could you set default=0 instead to remove the need for the `if default` check and the second `with socket.socket() as sock` codeblock. e.g something roughly like:

```
def get_free_port(default=0):
    with socket.socket() as sock:
        try:
            port = sock.bind(("0.0.0.0", int(default)))
        except OSError:
            # the socket is busy
            pass
        else:
            port = sock.getsockname()[1]
            return port
```

lucyb (approved):
One small suggestion, otherwise looks good and will be a useful feature.

opensafely/utils.py
bloodearnest: I think we need the 2nd socket check in case the default port is not free?

lucyb: Oh yes, I see. When I tried it locally the test was passing because it was returning None, which passes the assertion `assert utils.get_free_port(free_port) != free_port`.

How about this instead?
```
    with socket.socket() as sock:
        try:
            port = sock.bind(("0.0.0.0", int(default)))
        except OSError:
            # the socket is busy
            sock.bind(("127.0.0.1", 0))
        port = sock.getsockname()[1]
        return port
```

If it won't work that's fine, I'm just curious to see if it's possible to make the code a bit neater.

Actually, I don't think we need the assignment in the first `sock.bind` call, as that method always returns None. So:

```
port = sock.bind(("0.0.0.0", int(default)))
```
becomes:
```
sock.bind(("0.0.0.0", int(default)))
```

bloodearnest: That should work. I am not 100% is easier to understand, but it does have less branches. I will try it.

bloodearnest: nb we cannot bind to 127.0.0.1, as docker --port binds to 0.0.0.0 by default.

We could maybe change opensafely launch's --port argument to bind to 127.0.0.1 by default, which might be a good idea. However, it will need x-platform testing, as on win/mac, 127.0.0.1 is not always what you think it is, as docker on those platforms use a VM with its own IP.

PR #328, opened 2025-03-26
fix: ensure rstudio/python images exist before running launch command

bloodearnest:
If we let docker pull them lazily it causes two problems

- it can take so long that the open browser thread times out
- if the user ran in the background, then there is no output, and it
  looks like the command has hung

So, instead we ensure the relavnat image exists locally first, and if we
need to download it, we show that download output to the user

### DIFF
--- a/tests/test_launch.py
+++ b/tests/test_launch.py
@@ -33,7 +33,12 @@ def test_jupyter(run, no_user, monkeypatch, version):
     run.concurrent = True
 
     run.expect(["docker", "info"])
+    # test if already running
     run.expect(["docker", "inspect", "test_jupyter"], returncode=1)
+
+    run.expect(
+        ["docker", "image", "inspect", f"ghcr.io/opensafely-core/python:{used_version}"]
+    )
     run.expect(
         [
             "docker",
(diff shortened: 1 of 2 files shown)


tests/test_launch.py
```diff
@@ -69,8 +74,8 @@ def test_jupyter(run, no_user, monkeypatch, version):
     mock_open_browser.assert_called_with("http://localhost:1234/?token=TOKEN")
 
 
-@pytest.mark.parametrize("version", ["", "v2"])
-@pytest.mark.parametrize("gitconfig_exists", [True, False])
+@pytest.mark.parametrize("version", ["v2"])
+@pytest.mark.parametrize("gitconfig_exists", [False])
```
StevenMaude, line 78: question: What's the reason behind removing the arguments set by `parametrize` here? Or was this a development-only change that was committed?

(If we don't need these in the test, then the conditionals in the test that change behaviour according to them become redundant, and the `parametrize` itself becomes unnecessary.)

StevenMaude (approved):
I had one small question, but otherwise this looks fine to me.

PR #337, opened 2025-05-12
Custom UA for contacting opencodelists.org

### LINKED ISSUE #333: Provide a specific "opensafely cli" user-agent
We would like to enable Cloudflare's "bad bot" protections on OpenCodelists. However, when we [tried this previously](https://github.com/opensafely-core/opencodelists/issues/2217), we found that this blocked the OpenSAFELY CLI from connecting, because that shows up as `python/requests`, which Cloudflare deems to be a bot.

We should set a custom user-agent when [making requests](https://github.com/opensafely-core/opensafely-cli/blob/main/opensafely/codelists.py) to OpenCodelists.

This would also allow us to more easily track these requests in our Honeycomb dashboards.

Jongmassey:
Fixes #333 

We would like to enable Cloudflare's "bad bot" protections on OpenCodelists, `requests`'s standard UA of `python/requests` is flagged as a bot.

This PR introduces a bespoke user agent for OpenSAFELY CLI so that we can identify calls from the CLI

### DIFF
--- a/opensafely/codelists.py
+++ b/opensafely/codelists.py
@@ -12,6 +12,7 @@ from opensafely._vendor import requests
 
 OPENCODELISTS_BASE_URL = "https://www.opencodelists.org"
 DESCRIPTION = f"Commands for interacting with {OPENCODELISTS_BASE_URL}"
+HEADERS = {"User-Agent": "OpenSAFELY CLI"}
 
 CODELISTS_DIR = "codelists"
 CODELISTS_FILE = "codelists.txt"
@@ -149,7 +150,7 @@ def write_manifest(codelists_dir, downloaded_codelists, append):
 
 def fetch_codelist(codelist):
     try:
-        response = requests.get(codelist.download_url)
+        response = requests.get(codelist.download_url, headers=HEADERS)
         response.raise_for_status()
         content_type = response.headers["content-type"]
         if content_type != "text/csv":
@@ -207,7 +208,7 @@ def check_upstream(codelists_dir=None):
         "manifest": manifest_file.read_text(),
     }
     url = f"{OPENCODELISTS_BASE_URL}/api/v1/check/"
-    response = requests.post(url, post_data).json()
+    response = requests.post(url, post_data, headers=HEADERS).json()
     status = response["status"]
 
     if status == "error":
--- a/opensafely/codelists.py
+++ b/opensafely/codelists.py
@@ -7,16 +7,20 @@ import sys
 import tempfile
 from pathlib import Path
 
+import opensafely
 from opensafely._vendor import requests
 
 
 OPENCODELISTS_BASE_URL = "https://www.opencodelists.org"
 DESCRIPTION = f"Commands for interacting with {OPENCODELISTS_BASE_URL}"
-HEADERS = {"User-Agent": "OpenSAFELY CLI"}
 
 CODELISTS_DIR = "codelists"
 CODELISTS_FILE = "codelists.txt"
 MANIFEST_FILE = "codelists.json"
+
+
+def request_headers():
+    return {"User-Agent": f"OpenSAFELY-CLI/{opensafely.__version__.lstrip('v')}"}
 
 
 def add_arguments(parser):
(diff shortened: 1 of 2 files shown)


opensafely/codelists.py
```diff
@@ -18,6 +19,10 @@
 MANIFEST_FILE = "codelists.json"
 
 
+def request_headers():
+    return {"User-Agent": f"OpenSAFELY-CLI/{opensafely.__version__.lstrip('v')}"}
```
mikerkelly, line 23:
```suggestion
HEADERS =  {"User-Agent": f"OpenSAFELY-CLI/{opensafely.__version__.lstrip('v')}"}
```

I'm not sure if this needed to become a function rather than a module-level variable as it was before, which is a little simpler? I don't think the `opensafely` module can change version during the lifetime of this module, as it's only imported once.

tests/test_codelists.py
```diff
@@ -23,6 +23,7 @@ def _mock(response):
         mocked = requests_mock.post(
             "https://www.opencodelists.org/api/v1/check/",
             json=response,
+            request_headers=codelists.request_headers(),
         )
         return mocked
```
mikerkelly, line 28: I think it's a little non-obvious what this whole block is doing unless you're familiar with `requests_mock`. And in particular for this PR it's unclear that this validates the POST request headers. Worth a comment?

tests/test_codelists.py
```diff
@@ -438,3 +439,35 @@ def test_codelists_add_with_valid_non_codelist_url(
         )
     stdout, _ = capsys.readouterr()
     assert "No codelist found at URL" in stdout
+
+
+def test_codelists_update_user_agent(requests_mock, tmp_path):
```
mikerkelly, line 444: I think this test is almost redundant with `test_codelists_update` and doesn't need to be a separate test. I think adding the headers to the `request_mock.get` calls in that other test would be sufficient -- that's like adding a couple more asserts to that test.

If they stay as separate tests please add some brief comments that including the headers in the mock calls validates that they are included in the request.

mikerkelly (approved):

opensafely/codelists.py
Jongmassey: if it's a module-level variable a partially-initialised module `AttributeError` is raised. Wrapping it in a function that is called when everything is seemingly properly initialised (as per [`upgrade`](https://github.com/opensafely-core/opensafely-cli/blob/61c0e7411bdbf00d739b53e843e164dc3bb90dfc/opensafely/upgrade.py#L86C8-L89C9)) works

mikerkelly: Oh I see, this PR has introduced a circular import. `opensafely/__init__.py` imports `opensafely/codelists.py` which now imports `opensafely`. I suppose we should avoid that if we can do so easily enough.

Can the versioning stuff move out of `opensafely/__init__.py` into a separate module which then gets imported into `opensafely/__init__.py`, and `opensafely/codelists.py` import directly from that module, rather than circularly from `opensafely`?

Jongmassey: I'm inclined to put that refactoring into its own ticket

tests/test_codelists.py
Jongmassey: Yep, agree, one added. :+1:

tests/test_codelists.py
Jongmassey: >I think this test is almost redundant with test_codelists_update

I agree. While I was developing this PR it was convenient to have them separate but it makes sense to merge them as I did with the `check` tests. Now merged into one test.

Jongmassey:
Tested against opencodelists.org to make sure cloudflare etc. are happy with the proposed custom UA:

```python
>>> from pathlib import Path
>>> from opensafely.codelists import request_headers, fetch_codelist, check_upstream, Codelist
>>> headers = request_headers()
>>> headers
{'User-Agent': 'OpenSAFELY-CLI/99.99.99'}
>>> url = "https://www.opencodelists.org/codelist/user/jon_massey/shielding-patient-list-spl-flu-risk-based-codes/7f5612ce/"
>>> filename = Path("/tmp/codelist.csv")
>>> c= Codelist(
    id="user/jon_massey/shielding-patient-list-spl-flu-risk-based-codes/7f5612ce",
    version="7f5612ce",
    codelist="user/jon_massey/shielding-patient-list-spl-flu-risk-based-codes",
    url=url,
    download_url=f"{url}download.csv",
    filename=filename,
)
>>> fetch_codelist(c)
(datetime.datetime(2025, 5, 14, 18, 43, 13, 147824), '89081e6728c7e8f6134586155e5c2d6255c29918')
>>> filename.exists()
True
>>> filename.read_text()
# [... whole bunch of csv text as expected ...]
>>> check_upstream(Path("tests/fixtures/codelists"))
Codelists OK
True
```

PR #356, opened 2025-09-03
Spike: embed jobrunner@2.75.3 into opensafely-cli

### LINKED ISSUE #354: SPIKKE: formally embed old job-runner code base for local-run
Currently, job-runner is explicitly vendored using the `vendoring` tool, including all its dependencies. The last commit vendored this way was https://github.com/opensafely-core/job-runner/releases/tag/v2.75.3 in March.

At this point the work on the RAP API split started in job-runner main, and we basically hard forked the the local-run implementation.

This leaves us in the awkward position of not being able to practically make changes to the local_run behaviour, its frozen at that point.  opensafely-cli does not run the tests for it (it doesn't even have the test dependencies to do so).

This ticket is an attempt to spike the formally move the job-runner code base at that commit into opensafely-cli. It should be a first class citizen, its internal implementation of the `opensafely run` command. Importantly, opensafely cli test suite will need to be expanded to include the job-runner/local_run test suite, which will add some complexity.

Note that we still expect to use `vendoring` to manage the dependencies, as we don't want the `opensafely` pypi package to have external dependencies. But hopefully, the formal move will mean the import re-writes can go away, and that the normal python dependency update process should be work.

bloodearnest:
This is a mostly working spike to move the currently vendored jobrunnner@2.75.3
into opensafely-cli as a first class citizen.

Broadly speaking it:

 - still vendors the external dependencies, but not jobrunner
    - is much simpler and amenable to auto-updates.
 - adds a copy of jobrunner@2.75.3 as opensafely.jobrunner
 - adds a copy of the jobrunner test suite to tests/jobrunner
 - rewrites all the opensafely-cli imports to use the new package locations
 - rewrites all the jobrunner imports to import from opensafely.jobrunner
 - rewrites all the jobrunner external dependencies imports to use the vendored
   path

The original opensafely tests pass, and I have successful tested `opensafely
run` against some user code. Most of the jobrunner tests pass, there are about
30 failing, many of which related to code we want to delete.

Fixes #354


evansd (approved):
This is great. It must have been a right pain to do, but it gives us a solid base from which to move towards a nice, standalone local run solution.

The thoughtful commit sequence here made it a breeze to review, despite the enormous diff. In fact, this is a great example of where aiming for individual working commits would not be the right thing to do.

justfile
```diff
@@ -11,8 +11,8 @@ export COMPILE := BIN + "/pip-compile --allow-unsafe"
 
 
 # list available commands
-default:
-    @{{ just_executable() }} --list
+_default:
+    @just --list
```
StevenMaude, line 15: question: Why were these types of changes in the `justfile` required at this point? Were they required for making the Windows tests work, or are they incidental changes?

(Why I ask is that the repo template [currently uses `just_executable()`](https://github.com/opensafely-core/repo-template/blob/e9b08d4a8ce7836970955ea5309592b986bd72cf/justfile#L10-L11) and I assume that this feature of Justfiles works cross-platform.)

StevenMaude (approved):
I'll echo Dave's comments in saying that this is great, and that this large set of changes was made easy to review.

I think the approach that's been taken to move the code here is a reasonable one.

(I guess the possibility of removing/toggling telemetry mentioned in #354 is left for a future PR and to keep this smaller, rather than discounted entirely. In which case, we may want a fresh issue for that.)

justfile
bloodearnest: So, it doesn't seem to work cross platform.

On windows, `{{ just_executable() }}` returns `C:\path\to\just.exe`, i.e. with backslashes. `sh` understandably doesn't like this.

We know just is on the path in both linux and windows, so using it directly seems to work fine. In scenarios where just was wasn't on the path, this approach wouldn't work.

tests/__init__.py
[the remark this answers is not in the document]
bloodearnest: blah, yeah, missed a git add, will push

bloodearnest:
> (I guess the possibility of removing/toggling telemetry mentioned in #354 is left for a future PR and to keep this smaller, rather than discounted entirely. In which case, we may want a fresh issue for that.)

Yes, exactly that, once this lands I'll push that PR.

bloodearnest:
> suggestion: The explanations here are really helpful. I'd also suggest recording somewhere briefly in the commit message history:
> 
> * briefly why we decided on this change (it's in the PR message and associated linked threads, but not really in the commit messages)

The PR message ends up the merge commit, so it is recorded in the commit log?

> * why we are still vendoring (I think that's because we don't want to rely on local `pip` installs doing the right thing)

Good point, will add to the PR message.

justfile
remlapmot: If you want the original incantation to work maybe you can surround it in quotes so the single backslashes won't disappear on Windows, i.e.,
```just
default:
    @"{{ just_executable() }}" --list
```

bloodearnest: Yeah, I wondered if that may work, but I think I prefer using plain `just`.

{{ just_executable() }} is a solution to a problem I've not encountered yet.

bloodearnest: Also, I think it would need to be double backslashed \\ ?

StevenMaude:
> The PR message ends up the merge commit, so it is recorded in the commit log?

Dave mentioned this today too; I never knew this!

I think I just assumed that the message of the merge commit was "Merge pull request…".

evansd:
It turns out it's a [per-repo setting](https://bennettoxford.slack.com/archives/C01T2HACV3K/p1758102804959239?thread_ts=1758043120.215329&cid=C01T2HACV3K), which I never knew! I've obviously seen that behaviour in some repos and just assumed it was universal.

PR #360, opened 2025-09-23
Make otel dependencies optional

bloodearnest:
Make most of the otel dependencies optional, except for opentelemetry-api, which is still vendored. This removes them as required dependencies, but preserves the ability to use otel with `opensafely run` if needed. See #361 for more info.

Significant wins:
 - 9 less vendored packages, more than 50k lines of cod 
 - remove our pkg_resources hack
 - we can move back to upstream vendoring, and archive [our fork](https://github.com/opensafely-core/vendoring)
 - full integration tests that run a real job, with and without the optional dependencies

Fixes: #361


evansd (approved):
So nice seeing all this cruft disappearing. The new `trace()` abstraction looks like a nice simplification as well.

StevenMaude (approved):
This is a really good and well-explained simplification.

I didn't really know much about how OpenTelemetry is installed for Python. So it was useful to see the explanation around how the `opentelemetry-api` package can be still used without the other related `opentelemetry` dependencies installed. Leaving `opentelemetry-api` vendored then means we don't need to manage the availability of the other dependencies throughout our code, but only in a few places.

Related to that, a tiny nit in the PR message:

> Make most of the otel dependencies optional, except for opensafely-api, which is still vendored.

I think this should be `opentelemetry-api`.

PR #385, opened 2026-03-13
Bump pipeline version and add custom warning formatting

rebkwok:
The latest pipeline version introduces some warnings if a project.yaml is not using the latest version, or if it contains an action with the reserved "run_all" action name.
See https://github.com/opensafely-core/pipeline/pull/455

These warnings all start with "ProjectWarning:" and are expected to be printed to the console when a user runs `opensafely run`; any info other than the message itself (e.g. the line of code that produced it, the warning class etc) are
likely to be useless and confusing to users, so we just format them to display the message only. Other types of warning will continue to use the standard formatting.

Means that instead of output like:
```
➤ opensafely run run_all
/home/becky/bennett/opensafely-cli/opensafely/_vendor/pipeline/main.py:28: UserWarning: ProjectWarning: Your project file is using an old version (4.0); consider updating to version 5
  return Pipeline.build(**parsed_data)
/home/becky/bennett/opensafely-cli/opensafely/_vendor/pipeline/main.py:28: UserWarning: ProjectWarning: `run_all` is a reserved action name; user-defined actions with this name are ignored and will raise an error in later versions.
  return Pipeline.build(**parsed_data)
=> All actions already completed successfully
   Use -f option to force everything to re-run
```

We get: 
```
➤ opensafely run run_all
ProjectWarning: Your project file is using an old version (4.0); consider updating to version 5
ProjectWarning: `run_all` is a reserved action name; user-defined actions with this name are ignored and will raise an error in later versions.
=> All actions already completed successfully
   Use -f option to force everything to re-run
```

Also adds pytest config to ignore the pipeline warnings (and a driveby fix for missing marker registration, to silence the pytest unknown marker warnings)


DEVELOPERS.md
```diff
@@ -15,6 +15,12 @@ The tool makes the process relatively painless. There are a few
 workarounds (crude string subsitutions) we need to apply which are all
 configured in [pyproject.toml](./pyproject.toml).
 
+To update vendored dependecies, run:
+
+```
+./scripts/update.sh
```
StevenMaude, line 21: thought: Thanks for adding this note. Maybe it could be a `just` recipe instead.

tests/test_init.py
```diff
@@ -52,3 +55,29 @@ def test_warn_if_updates_needed_images_outdated(capsys, monkeypatch, tmp_path, r
         "    opensafely pull",
         "",
     ]
+
+
+def test_warnings_format():
```
StevenMaude: thought: Maybe it's worth a docstring/comment mention of what this is testing, that is, the patched behaviour? Or calling it `test_patched_format_warnings` or similar? Looking at the code, out of context, it might not be immediately apparent what's happening, and why this test is here.

opensafely/__init__.py
```diff
@@ -32,6 +33,26 @@
...
+    line=None,
+):
+    if str(message).startswith("ProjectWarning"):
+        return f"{message}\n"
+    return _original_formatwarning(message, category, filename, lineno, line)
+
+
+warnings.formatwarning = warning_message_only
+
+
```
StevenMaude, line 55: thought: This monkeypatching is not ideal, but I'm not sure there's another good simple way here.

StevenMaude (approved):

DEVELOPERS.md
rebkwok: I think it probably should, but I'm going to leave it for another PR because I'd like to check with other RAP team members first (e.g. I think it should have a dependency on `requirements-prod` and there's another commented out recipe in the justfile that probably just needs to be removed)

opensafely/__init__.py
rebkwok: no; but it seems to be a fairly standard way to do this

PR #386, opened 2026-03-13
Update Jobrunner local telemetry for consistency with production

rebkwok:
Remove the LOOP trace, which doesn't provide any useful info for local telemetry
Closes #382 

Update the tracing attributes to namespace with `job.` for consistency with production controller telemetry.
Closes #381

Latter changes largely follow [job-runner PR #1075](https://github.com/opensafely-core/job-runner/pull/1075)


bloodearnest (approved):
Thanks for this

PR #388, opened 2026-03-18
Add just recipes for updating vendored dependencies

rebkwok:
- add a copy of the upgrade-pipeline script used in other report
- move update vendored dependencies from script to a just recipe
- add recipe for updating pipeline which also updates vendored dependencies

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -15,11 +15,24 @@ The tool makes the process relatively painless. There are a few
 workarounds (crude string subsitutions) we need to apply which are all
 configured in [pyproject.toml](./pyproject.toml).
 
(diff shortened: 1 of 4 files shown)


bloodearnest (approved):
This makes sense to me, thanks.

It used to be more complex, and having it as an external script enabled shellcheck for it. But its pretty simple now, so I think embedding it as as just recipe makes sense.

The main reason I wrote update-pipeline.sh as separate script was to enable easily updating it across multiple repos.

But perhaps it could also be an embedded recipe?

rebkwok:
> But perhaps it could also be an embedded recipe?

I think it could, but I'll maybe leave that until we do it across all of them at once

PR #398, opened 2026-05-12
Remove `extract-stats` command

### LINKED ISSUE #394: Remove extract-stats command
This deprecated command was for cohortextractor, and should have been hidden from the public --help output, as was for internal usage.

However, it is still listed, and at the least should be hidden by default, and possibly removed alltogether

Image

StevenMaude:
Fixes #394.

Reasoning:

* It's already been deprecated a while.
* It was only for use with `cohort-extractor` logs, and that has now been marked as unsupported since June 2025; reference: https://github.com/opensafely-core/cohort-extractor/commit/bb89f630edc13d420934a15d6d86a360859e50ef


opensafely/extract_stats.py
KatieB5, line 1: Looking at the old script, I was wondering if the `.log` files in [tests/fixtures/metadata/old_style_logs](https://github.com/opensafely-core/opensafely-cli/tree/85e4c896e01246a6f005107e2be2f572ba46b5d1/tests/fixtures/metadata/old_style_logs) could also be removed as part of this PR? 

Maybe I've missed something but as far as I can tell, they were only used by the tests in `tests/test_extract_stats.py`, which is being removed in this PR.

KatieB5 (approved):
Thanks very much Steve!

opensafely/extract_stats.py
StevenMaude: Thanks; I'd missed that! Actually we can remove all of the `tests/fixtures/metadata` now as they're only used in the `extract-stats` tests 🧹 — so I'll rebase this with that in a single commit.

PR #402, opened 2026-05-25
Explain nondownloadable codelists

Mirochill:
## Summary
- Detect OpenCodelists' 'Codelist is not downloadable' response before the generic HTTP error path.
- Show users a clearer explanation that unsupported code column headers can make a codelist version unavailable for download.
- Add coverage for the nondownloadable codelist response.

Fixes #373

## Validation
- Not run locally.


lucyb:
Sorry, we aren’t currently accepting external contributions to our projects.

Issue #280, opened 2024-10-16
Drop support for Python 3.8


StevenMaude:
We test with Python 3.8, and also produce the build with Python 3.8. Python 3.8 is end-of-life as of 2024-10-07.

Existing users with local setups may be using an existing Python 3.8 installation to run this software. How do we get those users migrated? The solution there may be related to #277.

Are there other places where the Python version used for OpenSAFELY CLI interact with other components? (I *think* this should be independent, but I'll prompt the question in case not.)

bloodearnest:
Just to say, 3.9 is now EOL, so we should probably remove support for that too.

Additionally, this is keeping us from updating to the latest versions of many packages, as they no longer support 3.8 (or 3.9 in many cases).

Again, worth noting that moving to uv (#277) would be an alternative approach to solving this particular problem.

bloodearnest:
> > Stata, R or Python OpenSAFELY currently supports Stata v16.1, Python 3.8, Python 3.10, R 4.0 and R 4.4 for statistical analysis. For security reasons, available libraries are restricted to those provided by the framework.
> 
> [Introduction – Essential](https://docs.opensafely.org/#essential)

This is correct. This is talking about the python action image. `python:v1` will be python3.8 forever.

The version of python needed to run opensafely-cli is a separate thing.

> 
> > Python version 3.7 or higher must be installed on your machine to perform these tasks.
> 
> [Install python](https://docs.opensafely.org/install-python/)

This is out of date, and should be 3.8. 3.7 was the old minimum version for cohort extractor, I suspect its never been updated.

> 
> > `pyenv install 3.10:latest`
> 
> [Python](https://docs.opensafely.org/install-python/#python)

We should probably update this to track the most recent tested release of python for opensafely-cli, which is currently [3.13](https://github.com/opensafely-core/opensafely-cli/blob/main/.github/workflows/tests.yml#L10)

StevenMaude:
Though it makes things more complicated, we may also want to consider coinciding this change with #277. If users are forced to upgrade their installations, it might be good to also switch them to install with `uv`.

rebkwok:
Just adding to this - the pipeline library is also still on 3.8 due to its inclusion in opensafely-cli. It's now failing to apply a security update that isn't available on 3.8.

tomodwyer:
Blocked until we get a few more responses in the poll

PR #407, opened 2026-06-09
Drop support for legacy Python versions

tomodwyer:
Part of https://github.com/opensafely-core/opensafely-cli/issues/280 

This PR removes support in OpenSAFELY CLI for Python v3.8 and v3.9. As we are now updating to use 3.10, we take the opportunity to bump our dependencies, meaning we can also remove `black` and replace it with `ruff` - as we have in other repos.

This PR does not include `just update-vendored-dependencies` as that is a RAP task.

## Changes

- Bump minimum Python version to 3.10
- Update dependencies for Python 3.10+
- Replace black with Ruff
    Now that we have updated to a modern version of Ruff
- Run `just fix` to update syntax for Ruff


.github/workflows/tests.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-22.04, windows-2022]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
StevenMaude, line 16: suggestion: If some users are possibly upgrading their Python versions as a result of these changes, that may mean they upgrade to the latest stable version, which is 3.14.

So maybe we should change this to:

```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
```

and similarly add the same to the classifiers in `pyproject.toml`.

StevenMaude (approved):
The changes look fine; thanks for this.

Before we merge this, is it worth considering a final <3.10 supporting release with a check of the Python version at startup, and a warning if the Python version is <3.10?

The warning might state something like:

> The installed Python version being used to run the OpenSAFELY CLI is older than Python 3.10 and now end-of-life.
>
> This installed version of OpenSAFELY CLI will continue to run, but you will no longer receive new versions of the OpenSAFELY CLI.
>
> Please upgrade your Python installation to ensure you can upgrade to new versions of the OpenSAFELY CLI.

It's fine if the answer is "no", but I'm not sure if this idea has been discussed anywhere before.

StevenMaude:
thought: I would really like for us to go to further on this, but also concede that this is a reasonable and low risk improvement for now.

That's since:

* Python 3.10 is nearly five years old, which still restricts us from using some newer features.
* Python 3.10 is end-of-life as of end October 2026 (just three months). We'll be shortly in the situation again where dependencies can't be updated: they'll drop support for end-of-life Python.

But #280 was only scoped to the end-of-life versions incrementally.

So for now, I'll happily take a more regular maintenance cycle in line with Python end-of-life. Maybe other ideas like using `uv` (#277) make this simpler for end users to manage upgrades in future?

Doing this as a hopefully minimally disruptive upgrade with few users affected gets us some way towards modernising our users' installations, and figuring out how best to conduct these upgrade cycles.

.github/workflows/tests.yaml
tomodwyer: Great point, I'll update that.

tomodwyer: CI fails because of the version of OpenTelemetry required - I think we'll need to punt this until the dependencies are updated by RAP.

tomodwyer:
@StevenMaude yes I think moving the uv and getting off 3.10 before the end of its life cycle should be the next steps.

StevenMaude (approved):

tests/test_init.py
```diff
@@ -8,14 +8,22 @@
...
 
     assert "older than Python 3.10" in capsys.readouterr().err
 
 
+def test_warn_if_unsupported_new_python_version(capsys, monkeypatch):
+    monkeypatch.setattr(opensafely.sys, "version_info", (3, 14))
+
+    opensafely.warn_if_unsupported_python_version()
+
+    assert "newer than Python 3.13" in capsys.readouterr().err
```
StevenMaude: thought: You could probably `pytest.mark.parametrise()` this, combine it with the one above, and check for supported versions too. But the level of duplication is small, so I don't feel strongly.

For what it's worth, ChatGPT's attempt (untested, but looks plausible):

```python
@pytest.mark.parametrize(
    ("version_info", "expected_message"),
    [
        ((3, 9), "older than Python 3.10"),
        ((3, 10), None),
        ((3, 11), None),
        ((3, 12), None),
        ((3, 13), None),
        ((3, 14), "newer than Python 3.13"),
    ],
)
def test_warn_if_unsupported_python_version(
    capsys,
    monkeypatch,
    version_info,
    expected_message,
):
    monkeypatch.setattr(opensafely.sys, "version_info", version_info)

    opensafely.warn_if_unsupported_python_version()

    stderr = capsys.readouterr().err

    if expected_message is None:
        assert stderr == ""
    else:
        assert expected_message in stderr
```

tomodwyer: That is neat, I'll update and test it.

PR #410, opened 2026-07-08
fix: improve GitHub Actions security and permissions

tomodwyer:
In preparation for deprecating Python 3.8 and 3.9 - this PR covers off some fixes and updates to the GitHub Actions files, including:

- Using `.yml` file extension for all YAML files (as this was inconsistent)
- Pinning all GitHub Actions version numbers as hashes
- Fixing permissions and concurrency issues in Actions
- Formatting the YAML files with Prettier for consistency


.github/workflows/build_and_publish.yaml
```diff
@@ -1,22 +1,33 @@
 name: Tag repo; build and publish assets
+
 on:
   push:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
StevenMaude: question: Is the use of `pull_request` here due to testing? The trigger is only on `push`.

StevenMaude (commented):
Thanks!

.github/workflows/build_and_publish.yaml
tomodwyer: Yes, copy pasting from zizmor docs meant I forgot this only runs on main - will fix now.

StevenMaude (approved):

